PR #53, opened 2015-06-18
Enhanced location info

sakagg:
The parser, as of now only provides location information for a complete block, and not for the individual tags.
In my case needed location for both, start and end tags separately. Some other people might as well need it.
These are the minimum changes that I felt were required to make that happen. The tag locations can be accessed using __startTagLocation and __endTagLocation respectively.
Kindly review
Thanks!

### DIFF
--- a/lib/tree_construction/location_info_mixin.js
+++ b/lib/tree_construction/location_info_mixin.js
@@ -10,13 +10,15 @@ var $ = HTML.TAG_NAMES;
 
 
 function setEndLocation(element, endTagToken) {
-    if(!element.__startTagLocation && element.__location) {
+    if (!element.__startTagLocation && element.__location) {
         element.__startTagLocation = {
             start: element.__location.start,
             end: element.__location.end
         };
     }
-    if(endTagToken.location && endTagToken.type === Tokenizer.END_TAG_TOKEN && element.tagName === endTagToken.tagName) {
+    if (endTagToken.location &&
+        endTagToken.type === Tokenizer.END_TAG_TOKEN &&
+        element.tagName === endTagToken.tagName) {
         element.__endTagLocation = {
             start: endTagToken.location.start,
             end: endTagToken.location.end
@@ -108,7 +110,7 @@ exports.assign = function (parser) {
         //NOTE: _attachElementToTree is called from _appendElement, _insertElement and _insertTemplate methods.
         //So we will use token location stored in this methods for the element.
         element.__location = this.attachableElementLocation || null;
-        if(!element.__startTagLocation && element.__location) {
+        if (!element.__startTagLocation && element.__location) {
             element.__startTagLocation = {
                 start: element.__location.start,
                 end: element.__location.end
--- a/lib/tree_construction/location_info_mixin.js
+++ b/lib/tree_construction/location_info_mixin.js
@@ -10,8 +10,8 @@ var $ = HTML.TAG_NAMES;
 
 
 function setEndLocation(element, endTagToken) {
-    if (!element.__startTagLocation && element.__location) {
-        element.__startTagLocation = {
+    if (element.__location && !element.__location.startTag) {
+        element.__location.startTag = {
             start: element.__location.start,
             end: element.__location.end
         };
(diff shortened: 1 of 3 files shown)


lib/tree_construction/location_info_mixin.js
```diff
@@ -10,7 +10,19 @@ var $ = HTML.TAG_NAMES;
 
 
 function setEndLocation(element, endTagToken) {
-    if (element.__location)
+    if(!element.__startTagLocation && element.__location) {
```
davidbau: Add a space between `if (`.

lib/tree_construction/location_info_mixin.js
```diff
@@ -10,7 +10,19 @@ var $ = HTML.TAG_NAMES;
...
+        element.__startTagLocation = {
+            start: element.__location.start,
+            end: element.__location.end
+        };
+    }
+    if(endTagToken.location && endTagToken.type === Tokenizer.END_TAG_TOKEN && element.tagName === endTagToken.tagName) {
```
davidbau: Are the last two clauses needed, or are they already required to be true?

sakagg: Yes, both of them are required.
First one handles cases like
<code>&lt;p&gt; &lt;p&gt; &lt;/p&gt;</code>
and ensures no __endTagLocation for the first &lt;p&gt; (because it doesn't have one)

The second one handles cases like
<code>&lt;table&gt;</code>
<code>&nbsp;&nbsp;&lt;tr&gt;</code>
<code>&nbsp;&nbsp;&nbsp;&nbsp;&lt;td&gt;</code>
<code>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&lt;p&gt;</code>
<code>&nbsp;&nbsp;&nbsp;&nbsp;&lt;/td&gt;</code>
<code>&nbsp;&nbsp;&lt;/tr&gt;</code>
<code>&lt;/table&gt;</code>
and ensures no __endTagLocation for &lt;p&gt;

lib/tree_construction/location_info_mixin.js
sakagg: Done :+1:

lib/tree_construction/location_info_mixin.js
davidbau: Cool. Maybe add brief comments with those two cases next to the clauses that check against them.

davidbau:
Consider modifying the README.md to mention your new location field on the element.

inikulin:
Thank you for the contribution! The PR looks great, but we need to work through some minor issues: 
  1. Can we move tag location information into the `__location` object, so we will not pollute element with the element location properties? (E.g. `__startTagLocation` -> `__location.startTag`.

  2. We definitely need some tests for this.

davidbau:
A naming suggestion if merging things into a single object - maybe instead of calling it the `end` of the startTag, we should call it the start of the content, i.e., `start`, `contentStart`, `contentEnd`, `end`.

sakagg:
I think what inikulin mentioned is more intuitive as definition of 'content' can be ambiguous.
On the other hand daivdbau's method has less data repetition.
Not sure what to do.
Also, in cases where starting tags are missing but content is present, we can't set contentStart location due to lack of information of the startTag(which is a fake tag) which can be misleading.

inikulin:
I think `contentStart`, `contentEnd` is too generic and can be misleading. The following structure looks good for me:
```js
__location: {
    start: Number,
    end: Number,
    startTag: {
        start: Number,
        end: Number
    },
    endTag: {
        start: Number,
        end: Number
    }
}
```

Doctypes, comments and text nodes `__location` don't have `startTag` and `endTag` properties of course.

sakagg:
@inikulin @davidbau 
Updated!
Please have a look
Thank You

inikulin:
Thank you. LGTM.

inikulin:
I'll merge and roll out new version tomorrow.

sakagg:
Thanks! :smile:

inikulin:
Published as v1.5.0

sakagg:
Thanks :+1:

PR #56, opened 2015-07-21
allowing configuration of escaping entities

nathanhinish:
This will allow people to configure the string escape sequence to suit their needs without affecting the default behavior.

### DIFF
--- a/lib/serialization/serializer.js
+++ b/lib/serialization/serializer.js
@@ -11,7 +11,8 @@ var $ = HTML.TAG_NAMES,
 
 //Default serializer options
 var DEFAULT_OPTIONS = {
-    encodeHtmlEntities: true
+    encodeHtmlEntities: true,
+    escaper: escapeString
 };
 
 //Escaping regexes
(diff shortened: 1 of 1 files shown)


inikulin:
Use cases?

nathanhinish:
I'm using this module through jsdom. For my use, I needed ALL entities to be escaped again when serializing the document. This references go like this jsdom.serializeDocument -> domToHtml (in the jsdom module) -> parse5.TreeSerializer.

In order to do this currently, I am overwriting the _serializeAttributes and _serializeTextNode functions on the Serializer prototype. Then I am declaring my own escapeString function that uses the html-entities module to escape a MUCH larger set of entities.

I created a Gist to show my custom file. https://gist.github.com/nathanhinish/acd2ef056503d861f656

nathanhinish:
And here's an HTML file that lists all the encoded entities that I used initially to see what was happening in my code.

https://gist.github.com/nathanhinish/c8ac4d37f2fb89f121bd

inikulin:
I don't think we need this function in parse5, since you can just walk the tree, escape required text node and then serialize with the  disabled `encodeHtmlEntities` option.

PR #62, opened 2015-08-25
New streaming API

inikulin:


aredridel:
This is super exciting! That's one of the few things `html5` does that this didn't.

fb55:
@inikulin Awesome progress. We should probably revisit the topic of merging whacko with cheerio at some point :)

inikulin:
@fb55 I'd love to do that. Please, let me know if you need any assistance from me.

Issue #31, opened 2014-12-16
Provide options to disable HTML entities processing in Parser/Serializer


alanclarke:
it would be cool if parse5's serializer had an option to preserve the provided html without encoding html entities

PR #74, opened 2015-09-30
grammar fix in api docs

stevenvachon:
Minor change. I think that a link to a wiki page/etc explaining just how and what non-conformities can occur would be helpful. I'm even wondering as you read this.

### DIFF
--- a/lib/parser/index.js
+++ b/lib/parser/index.js
@@ -21,7 +21,7 @@ var $ = HTML.TAG_NAMES,
  * @typedef {Object} ParserOptions
  *
  * @property {Boolean} [decodeHtmlEntities=true] - Decode HTML-entities like `&amp;`, `&nbsp;`, etc.
- * **Warning:** disabling this option may cause output which is not conform HTML5 specification.
+ * **Warning:** disabling this option may result in output that does not conform to the HTML5 specification.
  *
  * @property {Boolean} [locationInfo=false] - Enables source code location information for the nodes.
  * When enabled, each node (except root node) has `__location` property. In case the node is not an empty element,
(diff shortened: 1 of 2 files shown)


inikulin:
Thank you, I'm currently switching to the jsdoc auto-generated docs. So, fix appropriate doc comments instead of README, please: 
 * https://github.com/inikulin/parse5/blob/master/lib/parser/index.js#L23
 * https://github.com/inikulin/parse5/blob/master/lib/serializer/index.js#L16

>  I think that a link to a wiki page/etc explaining just how and what non-conformities can occur would be helpful.

Well, it will obviously not parse/serialize HTML entities how spec suggests.

stevenvachon:
Obvious to whom? Not to me.

Corrections made.

inikulin:
Gosh, that's why I hate all this optional spec violations and other weird edge case things: it's not enough to tell people what they can mess up everything, you should describe in greater detail what could possibly go wrong. And nearly as always the answer is: **anything, except your edge case scenario for which we've implemented this options**.

So the answer is: anything, except edge case scenarios for which we've implemented this options (#31). E.g.: running serializer with `encodeHtmlEntities` enabled on the output of parser with the `decodeHtmlEntities` disabled: `&amp;` will be serialized as `&amp;amp;`

inikulin:
Thank you!

stevenvachon:
Dude, we don't all write and maintain spec-compliant HTML parsers. If we did, you wouldn't need to write one.

The browser decodes and encodes entities within attributes. It probably does so for text content as well so that it can render, but web inspectors re-encode them for easy reading.

With this new knowledge, I agree that `decodeHtmlEntities` should probably be removed. A note regarding manual re-encoding would be helpful to those looking to skip decoding.

PR #80, opened 2015-12-07
documentation fix

stevenvachon:


stevenvachon:
With this, is it possible to recreate functionality from v1.5 where we fed a string into the SAX parser?:

inikulin:
Thanks!

> With this, is it possible to recreate functionality from v1.5 where we fed a string into the SAX parser?:

Sure:
```js
var parser = new parse5.SAXParser();
// ..handlers

parser.end(str);
```

stevenvachon:
Ah, thanks. There is no documentation for `SAXParser.prototype.end`, by the way.

inikulin:
@stevenvachon It's [Stream](https://nodejs.org/api/stream.html) API. And there is a link to it in the docs: https://github.com/inikulin/parse5/wiki/Documentation#new-saxparseroptions. Copying node docs  here doesn't make much sense.

PR #86, opened 2015-12-12
Fix couple of mistakes in TreeAdapter API docs

RReverser:
(There is something weird going on with Github diff - the actual changed lines are not the ones it shows)

PR #87, opened 2015-12-15
Fix <template> references in docs which caused table corruption.

RReverser:


domenic:
Not sure but I think these docs are auto-generated so you need to edit the .js files instead.

RReverser:
@domenic Also not sure about this part, but #86 was accepted by @inikulin where I edited docs manually.

If they are auto-generated, then I'm somewhat confused why `<` and `>` are correctly replaced in `ASTNode<DocumentFragment>` but not in `<template>` and not sure where to fix :/

inikulin:
@RReverser My bad, @domenic is right. You need to edit jsdoc comments in the code, since whole API reference section is auto-generated. I think, I'll gitignore `docs/05_api_reference.md`to avoid confusion in the future. Regarding #86 didn't looked md-file diff, since jsdoc randomly moves members across generations. So, I'm afraid you will need to apply this changes as well. Sorry for that.

RReverser:
@inikulin No worries. Doesn't explain my question though - shouldn't generator handle that as it does in `ASTNode`? Don't really want to put `&lt;` / `&gt;` entities in real code comments just for the sake of generated Markdown :(

inikulin:
@RReverser to be honest, I have no idea. jsdoc generator is extremely buggy and even doesn't work for me on OSX :cry:

inikulin:
@RReverser I'm trying to figure out the whole issue to answer your question, but looking at the diffs - they look identical for me (or my vision is too blured after working day :tired_face:). Can you explain what's wrong currently, please? Maybe I'll have some ideas.

RReverser:
@inikulin Same issue as in #86, apparently Github doesn't recognize specific line endings generated by jsdoc or smth :(

> What's even more weird, while RReverser@c2f8f4f shows incorrect diff, https://github.com/RReverser/parse5/commit/c2f8f4f538aa6c6ed60ebc38e8c28af60f11f6d7.diff returns the correct one. Apparently there is some bug with Github.

Same here - just add `.diff` or `.patch` to URL to see the correct raw diff. The only reliable workaround that I've found so far.

inikulin:
@RReverser got it now. Well, seems like it's GH's markdown sanitizer - it treats `<template>` as the forbidden HTML tag. 
Quick workaround: wrap tag names with "`" in jsdoc comments, e.g.:
```
Sets the `<template>` element content element.
```
AFAIR, I already did it in some places.

inikulin:
Fixed in fc4c510def57e75845c381257b55ca8abb62caba

RReverser:
Thanks! Sorry, didn't find time to finish.

inikulin:
@RReverser It's ok, thank you for pointing this out and taking time to fix it?

PR #88, opened 2015-12-15
Add link to parse5 playground at ASTExplorer

RReverser:
This should allow people to get better sense of what `parse5` AST looks like.


inikulin:
This is awesome, thank you!

PR #91, opened 2016-01-04
Fix parseFragment with locationInfo regression when parsing <template…

### LINKED ISSUE #90: regression: parseFragment with locationInfo: true on <template>
When using `parse5.parseFragment()` with `locationInfo: true`, and the fragment contains `<template>` element, the location info parser mixin throws an error here: https://github.com/inikulin/parse5/blob/master/lib/location_info/parser_mixin.js#L163

This is due to the 2.0 breaking change in the default tree adaptor where the `<template>` node now holds its content in `content` instead of in `childNodes`.

yyx990803:
…> (fix #90)

Fixes #90. Introduces a new method for tree adaptors: `getTemplateContent`.

### DIFF
--- a/lib/location_info/parser_mixin.js
+++ b/lib/location_info/parser_mixin.js
@@ -158,7 +158,7 @@ exports.assign = function (parser) {
         attachableElementLocation = token.location;
         parserProto._insertTemplate.call(this, token);
 
-        var tmplContent = this.treeAdapter.getChildNodes(this.openElements.current)[0];
+        var tmplContent = this.treeAdapter.getTemplateContent(this.openElements.current);
 
         tmplContent.__location = null;
     };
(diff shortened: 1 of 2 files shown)


lib/tree_adapters/default.js
```diff
@@ -512,6 +512,23 @@ exports.getDocumentTypeNodeSystemId = function (doctypeNode) {
...
+ * @function getTemplateContent
+ * @memberof TreeAdapter
+ *
+ * @param {ASTNode} node - Node.
+ *
+ * @returns {ASTNode}
+ *
+ * @see {@link https://github.com/inikulin/parse5/blob/tree-adapter-docs-rev/lib/tree_adapters/default.js#L527|default implementation.}
+ */
+exports.getTemplateContent = function (node) {
```
inikulin: No need for this, it's already implemeted: https://github.com/yyx990803/parse5/blob/master/lib/tree_adapters/default.js#L166

inikulin:
Thank you for finding and fixing it! Just a little issues with some excessive code =_

yyx990803:
Haha, I didn't even check... updated.

inikulin:
lgtm, thanks!

PR #95, opened 2016-01-13
fix attribute location info for empty attributes

yyx990803:
In my PR for attribute location I forgot to also overwrite `tokenizer._leaveAttrName` - which causes missing location info for attributes with no value.

For example, `<div some-attr></div>` would have an `attrs` array of length 1, but no `__location.attrs`.


inikulin:
Can you add regression test please?

yyx990803:
Added test case.

inikulin:
Thanks!

inikulin:
Published as v2.1.1

Issue #99, opened 2016-01-26
Foreign content attributes and tag names are not adjusted by the SAXParser


inikulin:

pixelami:
I have noticed that when using the SAXParser to process a ReadableStream instance (containing latin and non latin characters). I notice the `location` objects passed to the 'startTag' and 'endTag' event handlers have incorrect `startOffset` and `endOffset` positions.

Does that seem to be a likely symptom of this issue of should I create a separate issue and investigate further?

inikulin:
@pixelami Seems like it's completely different issue

pixelami:
I created this issue,
https://github.com/inikulin/parse5/issues/103

Is it feature or a bug, I'll let you decide....

PR #127, opened 2016-05-11
Parser feedback

### LINKED ISSUE #115: Test parser feedback simulator for SAXParser
https://twitter.com/_inikulin_/status/715621108809076737

RReverser:
As of the moment of committing, 1482 tests are passing and 66 are failing. Some of those we can easily fix, but some not as they would require infinite buffering otherwise.

Closes #115.


inikulin:
LGTM. I guess we will keep this PR open for now, until we will not fix failing tests and figure out what to do with those that we don't know how to fix.

inikulin:
I see lots of fails caused by #99, shouldn't be hard to fix

RReverser:
UPDATE: 55 failing now; made `adjustTokenizer` private (`_adjustTokenizer`) and instead exposed `getNextToken` wrapper to be able to skip / substitute tokens more easily.

RReverser:
> I see lots of fails caused by #99, shouldn't be hard to fix

Sure, that's one of those that I referred to as

> Some of those we can easily fix

(another was first `\n` in `pre` / `textarea`)

RReverser:
Done, 28 failing left. Working on it :)

lib/sax/parser_feedback_simulator.js
```diff
@@ -20,10 +20,15 @@ var ParserFeedbackSimulator = module.exports = function (tokenizer) {
...
+    this.skipNextNewLine = false;
+};
+
+ParserFeedbackSimulator.prototype.getNextToken = function () {
+    return this._adjustTokenizer(this.tokenizer.getNextToken());
 };
 
 //API
-ParserFeedbackSimulator.prototype.adjustTokenizer = function (token) {
+ParserFeedbackSimulator.prototype._adjustTokenizer = function (token) {
```
inikulin: Maybe it makes sense to rename it to `_handleToken` now?

inikulin:
Regarding `ignored` token's property - it's parser intrinsic artifact and can be ignored during comparison, so we can easily fix yet another bunch of tests

lib/sax/parser_feedback_simulator.js
RReverser: Yeah, maybe. Or could even inline into this new `getNextToken` completely.

RReverser:
@inikulin Can you elaborate on this `ignored` thing? What is/was it used for?

inikulin:
In earlier versions of spec some end tag tokens were handled in the following manner:
- Create fake token
- Reprocess token in state ...
- If token was ignored in state, do ...

However, seems like someone clarified this sections because I can't find such instructions in the recent version of the spec (cool cool).  Both regular and fake tokens has same set of properties to avoid bailouts in handlers. So:
- it has nothing to do with tokenization (so we can don't care about it in feedback simulator)
- we can completely get rid of this property by updating appropriate handlers in parser per recent spec

RReverser:
> we can completely get rid of this property by updating appropriate handlers in parser per recent spec

That would be nicer than just hacking in order to ignore them in a specific test suite.

inikulin:
> That would be nicer than just hacking in order to ignore them in a specific test suite.

I agree

lib/sax/parser_feedback_simulator.js
```diff
@@ -76,13 +89,31 @@ ParserFeedbackSimulator.prototype._ensureTokenizerMode = function (tn) {
...
         this._enterNamespace(NS.SVG);
+        foreignContent.adjustTokenSVGAttrs(token);
+        foreignContent.adjustTokenXMLAttrs(token);
+    }
 
     else if (tn === $.MATH)
         this._enterNamespace(NS.MATHML);
 
+    else if (tn === $.IMAGE)
+        token.tagName = $.IMG;
```
inikulin: I wonder if we really should do this and new lines adjustment for `<textarea>` and stuff? I thought we should be close to the source code as much as possible, so maybe we should just ignore those tests? However, it's debatable. What do you think?

RReverser: Well, for our use case I need to be as close to browser behavior as possible, as I'm going to do various streaming transformations, so if browser treats `<image />` as `` then I should be able to see it this way too, so that any transformations that are applicable to `img` will be applied to `image` as well, any selectors that look for `img`, should fine `image` too, and so on and so on.

As for the newlines - that might be debatable, but, personally, I'd prefer to be as consistent with the real handling as possible in order not to decrease risk of running into weird bugs in future.

With the "close to the source code as much as possible" you could argue that we can also disable NUL-character and CRLF normalizations as they might also be seen as non-critical (especially in JS land), but that's a way too big rabbit hole to get into IMO :)

RReverser: Btw, theoretically, if we get close enough, we could even use this simulator in between the real parser and tokenizer in order to avoid logic duplication and share more code between non- and streaming parsers.

RReverser:
13 failing left. Getting there 😄

RReverser:
Okay, I added special handling for testing HTML / BODY, and added a note about why we're doing this. I guess this is the best way to handle this without breaking assumptions nor ignoring those tests completely.

RReverser:
Foreign content is a mess. Stuck on fixing those last 4 :/

RReverser:
Okay, *looks like* I've finally fixed it, but this last commit involved almost random reordering of some block until it worked, so it would be nice to have a review 😄 

In the other news, entire test suite now passes.

lib/parser/index.js
```diff
@@ -28,7 +28,8 @@ var $ = HTML.TAG_NAMES,
  */
 var DEFAULT_OPTIONS = {
     locationInfo: false,
-    treeAdapter: defaultTreeAdapter
+    treeAdapter: defaultTreeAdapter,
+    onToken: function () {}
```
inikulin: Can you add a comment that we expose it just for the testing purposes, please?

lib/sax/parser_feedback_simulator.js
```diff
@@ -82,17 +93,41 @@ ParserFeedbackSimulator.prototype._handleStartTagToken = function (token) {
...
+        else if (currentNs === NS.SVG) {
+            foreignContent.adjustTokenSVGTagName(token);
+            foreignContent.adjustTokenSVGAttrs(token);
+        }
+
+        foreignContent.adjustTokenXMLAttrs(token);
+
+        tn = token.tagName;
+
+        if (!token.selfClosing && (
```
inikulin: Maybe extract this condition into boolean var to get rid of this weird formatting and give it some context?

inikulin:
Great job! Just some minor style issues left.

lib/sax/parser_feedback_simulator.js
RReverser: Sounds good to me. Given that exact same combined condition is used in more than one place, perhaps better to create additional helper method on `foreignContent` itself?

inikulin: Yes, makes sense

lib/parser/index.js
RReverser: Initially, I actually hooked into private methods of `Parser` in the test suite, but then decided that it might be not the best idea and consumers might want to perform own actions on token retrieval as well (similarly to `treeAdapter`, but token-based). If you think exposing it is a bad idea and we should only use it for testing, then I think internal hook into `Parser` private methods would be better as it doesn't expose anything.

inikulin: >  consumers might want to perform own actions on token retrieval

Do you have any scenario in mind?

RReverser: Didn't really have any, that's why said "might" :) I'm totally fine with hiding it back if you wish.

inikulin: Let's hide it then, if someone will be interested in this feature we can always bring it back.

RReverser:
Ok done.

inikulin:
lgtm, thanks!!

PR #128, opened 2016-05-12
Avoid some code duplication

RReverser:
Since the related spec editorial changes have landed and we've seen no evidence of performance degradation in parse5 while it does simplify some code, decided to send those local commits as a PR.

### DIFF
--- a/lib/location_info/tokenizer_mixin.js
+++ b/lib/location_info/tokenizer_mixin.js
@@ -82,8 +82,8 @@ exports.assign = function (tokenizer) {
         attachLocationInfo(this.currentToken);
     };
 
-    tokenizer._createDoctypeToken = function (doctypeNameFirstCh) {
-        tokenizerProto._createDoctypeToken.call(this, doctypeNameFirstCh);
+    tokenizer._createDoctypeToken = function (initializeName) {
+        tokenizerProto._createDoctypeToken.call(this, initializeName);
         attachLocationInfo(this.currentToken);
     };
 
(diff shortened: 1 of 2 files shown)


RReverser:
Correction: DOCTYPE-related changes were not landed on spec due to changed `parse error` behavior which I yet have to fix, but since we don't care about parse errors, this should be fine already.

lib/location_info/tokenizer_mixin.js
```diff
@@ -82,8 +82,8 @@ exports.assign = function (tokenizer) {
         attachLocationInfo(this.currentToken);
     };
 
-    tokenizer._createDoctypeToken = function (doctypeNameFirstCh) {
-        tokenizerProto._createDoctypeToken.call(this, doctypeNameFirstCh);
+    tokenizer._createDoctypeToken = function (initializeName) {
```
inikulin: `initialName`? Having verb in the var name suggests that it's a function

inikulin:
Everything else - LGTM

lib/location_info/tokenizer_mixin.js
RReverser: Hm, yeah, no idea why I wrote it that way back then.

RReverser: Ah, it's because it's a boolean and `initialName` would be misleading as it could suggest it's a string.

RReverser: What should I change it to in such case?

inikulin: `shouldInitializeName`?

RReverser: Nah, just changed it to accept `initialName` - that's a bit cleaner anyway IMO 😄

inikulin:
lgtm, thanks!

PR #138, opened 2016-06-14
Generate feedback tests

RReverser:
Adds pre-generated parser feedback tests in HTML5Lib format and a task for generation of parser feedback tests in HTML5Lib tokenization format (and a task for updating them), so that they can be readable and reusable instead of generating them in runtime.

Since it's now in the same format as regular tokenization tests, merged parser_feedback_test logic into tokenizer_test (this required only extra parameter to be passed around the runner).

Additionally this helped to test parser feedback with chunked input and identify & fix an error related to extraneous newlines after hibernation.


inikulin:
Awesome (just as always)! Thanks!

RReverser:
Oh, you squashed them. I actually left a note why I did two separate commits there - Github seems to have problems with displaying such a huge diff (it hangs a browser), so left them as separate ones. Although probably it doesn't matter as long as nobody tries to see what has changed :)

PR #145, opened 2016-08-03
Type check str param in escapeString method

visikov:
There was an issue where false | undefined | null were being passed into parse5 via angular2-universal. That caused `TypeError: str.replace is not a function` errors.


RReverser:
Hm, I would say this is a problem of a consumer that passes incorrect values.

inikulin:
Hi, thank you for the contribution, but I agree with @RReverser here. We expect correctly formed as an input.

PR #146, opened 2016-08-04
Apply latest spec changes

inikulin:
@RReverser Can you review, please?


RReverser:
Omg :) I'll do tomorrow if you don't mind, that's lots of changes as for the evening.

lib/parser/index.js
```diff
@@ -950,30 +894,36 @@ function aaInnerLoop(p, furthestBlock, formattingElement) {
         lastElement = furthestBlock,
         nextElement = p.openElements.getCommonAncestor(furthestBlock);
 
-    for (var i = 0; i < AA_INNER_LOOP_ITER; i++) {
+    for (var i = 0; ; i++) {
```
RReverser: This might look not very clean IMO because logic of stop condition is hidden inside of the body. Perhaps better to put condition here explicitly? If it's not dependant on `i`, then at least `nextElement !== formattingElement` would do.

lib/parser/open_element_stack.js
```diff
@@ -278,12 +277,13 @@ OpenElementStack.prototype.hasInScope = function (tagName) {
 
 OpenElementStack.prototype.hasNumberedHeaderInScope = function () {
     for (var i = this.stackTop; i >= 0; i--) {
-        var tn = this.treeAdapter.getTagName(this.items[i]);
+        var tn = this.treeAdapter.getTagName(this.items[i]),
+            ns = this.treeAdapter.getNamespaceURI(this.items[i]);
 
-        if (tn === $.H1 || tn === $.H2 || tn === $.H3 || tn === $.H4 || tn === $.H5 || tn === $.H6)
+        if ((tn === $.H1 || tn === $.H2 || tn === $.H3 || tn === $.H4 || tn === $.H5 || tn === $.H6) && ns === NS.HTML)
```
RReverser: IIUC, those functions are pretty cold, and any tree adapter implementation are assumed to be fast, too, but otherwise I would probably split this into two `if`s where external one calls only tag name, and if it matches, only then calls `getNamespaceURI` to check it too, thus creating a "fast-case" for any elements in stack that are not headers.

But, as I said, might be not beneficial in this particular case, dunno.

lib/parser/index.js
```diff
@@ -520,6 +520,13 @@ Parser.prototype._insertElement = function (token, namespaceURI) {
     this.openElements.push(element);
 };
 
+Parser.prototype._insertFakeElement = function (tagName) {
+    var element = this.treeAdapter.createElement(tagName, NS.HTML, []);
+
+    this._attachElementToTree(element);
```
RReverser: Wondering if this should be made conditional (depending on some option) to address cases where people don't want fake elements inserted.

RReverser:
Awesome work! Added some notes, but LGTM.

lib/parser/open_element_stack.js
inikulin: > elements in stack that are not headers 

If elements are not headers we will need namespace anyway to perform scoping element test: https://github.com/inikulin/parse5/pull/146/files#diff-c84848d80bdda10576c7ee69b4b5f652R295

lib/parser/index.js
inikulin: As far as I remember there is only one such case: for `</br>`. But it's can be easily solved by immediately poping element out of the stack. For all other cases we need insertion and I believe we shouldn't introduce additional condition for this.

inikulin:
Thank you for the notes! Fixed AAA loop condition.

inikulin:
@RReverser Feel free to merge if you don't have any additional remarks

lib/parser/open_element_stack.js
RReverser: Ah right.

RReverser:
@inikulin Just back from the vacation, but good that you merged it already :)

zcorpan:
FYI https://github.com/whatwg/html/issues/2308

PR #151, opened 2016-08-31
Fix line number calculation bug

ChadKillingsworth:
When consuming tokens in script tags, a `<` right before a newline would cause the line number calculation to be offset by 1. Repo case:

```html
<!DOCTYPE html>
<html>
<body>
<script>
  var x = window.scrollY <
      100; //line number is now off by one
</script>
</html>
```

Simply setting the `isEol` flag to false when unconsuming a token corrects the issue.

### DIFF
--- a/lib/location_info/tokenizer_mixin.js
+++ b/lib/location_info/tokenizer_mixin.js
@@ -57,6 +57,7 @@ exports.assign = function (tokenizer) {
 
     tokenizer._unconsume = function () {
         tokenizerProto._unconsume.call(this);
+        isEol = false;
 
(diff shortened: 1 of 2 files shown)


inikulin:
Nice catch, thank you! Can you add regression test, please?

ChadKillingsworth:
Done.

inikulin:
Awesome, thank you!

inikulin:
Published as `v2.2.1`.

meisl:
TODO:
* did this bug materialize already at the tokenization level? Regression test here is for parser only...
* if so, add appropriate regression test to `packages/parse5/test/location-info-tokenizer.test.js` (see #284)

PR #154, opened 2016-09-26
Implement nearly constant memory consumption for huge streaming data by dropping unused buffer chunks (fixes #130, fixes #102)

### LINKED ISSUE #102: Parse5.SaxParser vs HtmlParser2: performance and memory issues
I made two testbenches for **HtmlParser2** and **Parse5.SaxParser**

### HtmlParser2
https://gist.github.com/ivan-kleshnin/fcf6353242f3a4a462e7

### Parse5.SaxParser
https://gist.github.com/ivan-kleshnin/bb571f51d47aaebee16a

File `huge.html` is **112mb** file containing **120472** paragraphs.
I uploaded it [here](http://www.filedropper.com/huge)  but you can easily make something similar.

---

Results are quite shocking :fearful: 

### HtmlParser2
```
stopped (4.41 seconds)
Mem before: 18.68 mb
Mem after: 16.32 mb
Mem max: 0b
```

### Parse5.SaxParser
```
stopped (1.62 minutes)
Mem before: 19.63 mb
Mem after: 125.46 mb
Mem max: 1.04 gb
```

There should be some bug explaining that. I would simply use **HtmlParser2** 
but I like **Parse5** API better so I hope you will fix that.

### LINKED ISSUE #130: Streaming implementation buffers text infinitely
If chunk has ended in DATA / RCDATA / ... states, it's possible to emit most of the collected text (before the entity or smth else) without buffering it between chunks. Currently, parse5 attempts to collect all the text into the internal buffer until it finds a beginning of the new token (like a starting tag), and only then emits altogether which is highly inefficient for documents containing long text chunks.

inikulin:
**Before:**
```js
Input data size: 183 MB
Duration:  31.16 seconds
Memory before:  25.57 mb
Memory after:  385.05 mb
Memory max:  1.43 GB
```

**After:**
```js
Input data size: 183 MB
Duration:  16.02 seconds
Memory before:  25.62 mb
Memory after:  25.04 mb
Memory max:  53.13 MB
```

Most of `53.13 MB` is a benchmark-related stuff. In original benchmark from #102 we now have ~10Mb memory consumption, which is primarily entities trie.

@RReverser, can you review, please?


inikulin:
Arghhh, for some reason node `v6.6.0` can't install native dependency required for benchmark. I'll make it optional

inikulin:
Fixed

inikulin:
More results on bigger data:
```
```

RReverser:
Sure, will take a look today!

RReverser:
What about speed benchmarks? Given that you call that method for each character now, curious if you noticed any slowdowns.

inikulin:
@RReverser 
Surprisingly, it doesn't affect performance much. The difference is within an order of statistical fluctuations:  

```
[16:07:11] Starting 'benchmark'...
[16:07:11]   Running 'parse5 regression benchmark - HUGE' 
[16:07:18]     Working copy x 6.53 ops/sec ±9.22% (21 runs sampled)
[16:07:24]     Upstream x 6.52 ops/sec ±6.73% (21 runs sampled)
[16:07:24]   'parse5 regression benchmark - HUGE' (passed: 2, failed: 0)
[16:07:24]     Passed:
[16:07:24]       'Working copy' at 1.00x faster
[16:07:24]       'Upstream' is etalon
[16:07:25]   Running 'parse5 regression benchmark - MICRO'
[16:07:30]     Working copy x 40.29 ops/sec ±9.80% (58 runs sampled)
[16:07:36]     Upstream x 39.47 ops/sec ±5.95% (54 runs sampled)
[16:07:36]   'parse5 regression benchmark - MICRO'  (passed: 2, failed: 0)
[16:07:36]     Passed:
[16:07:36]       'Working copy' at 1.02x faster
[16:07:36]       'Upstream' is etalon
[16:07:36]   Running 'parse5 regression benchmark - PAGES'
[16:07:41]     Working copy x 134 ops/sec ±4.14% (56 runs sampled)
[16:07:47]     Upstream x 137 ops/sec ±3.88% (57 runs sampled)
[16:07:47]   'parse5 regression benchmark - PAGES' (passed: 2, failed: 0)
[16:07:47]     Passed:
[16:07:47]       'Upstream' is etalon
[16:07:47]       'Working copy' at 1.02x slower
[16:07:47]   Running 'parse5 regression benchmark - STREAM' 
[16:07:53]     Working copy x 102 ops/sec ±4.36% (73 runs sampled)
[16:07:59]     Upstream x 104 ops/sec ±5.18% (59 runs sampled)
[16:07:59]   'parse5 regression benchmark - STREAM'  (passed: 2, failed: 0)
[16:07:59]     Passed:
[16:07:59]       'Upstream' is etalon
[16:07:59]       'Working copy' at 1.03x slower
[16:07:59] Finished 'benchmark' after 48 s
```

RReverser:
@inikulin Cool! Waiting for answers on review questions :)

inikulin:
@RReverser 

> Waiting for answers on review questions :)

I can't see any. Are you sure you've submitted the review?

inikulin:
ping @RReverser

RReverser:
Indeed, they're in comments (as in this new review process).

package.json
```diff
@@ -23,7 +23,8 @@
     "gulp-jsdoc-to-markdown": "^1.1.1",
     "gulp-mocha": "^2.1.3",
     "gulp-rename": "^1.2.2",
-    "promise": "^7.0.4",
+    "path-exists": "^3.0.0",
+    "pinkie": "^2.0.4",
```
inikulin: Nah, I would like to keep node for now, since parse5 is still used in some browsers without proper ES6 support, also node 10 is still widely used. I would like to make sure that we don't break anything by using ES6 syntax

RReverser: >  I would like to make sure that we don't break anything by using ES6 syntax

That sounds more like a task for linter.

inikulin: @RReverser Do you know any eslint plugin that restricts ES6 features usage?

RReverser: AFAIK just setting `ecmaVersion` to `5` should be enough (ESLint / Espree uses Acorn under the hood, so when it passes `ecmaVersion: 5`, Acorn simply won't recognize ES6 syntax and won't be able to parse it).

lib/tokenizer/preprocessor.js
```diff
@@ -26,15 +31,34 @@ var Preprocessor = module.exports = function () {
...
 };
 
+Object.defineProperty(Preprocessor.prototype, 'sourcePos', {
+    get: function () {
+        return this.droppedBufferSize + this.pos;
+    }
+});
+
+Preprocessor.prototype.dropParsedChunk = function () {
+    if (this.pos > this.bufferWaterline) {
```
RReverser: I meant calling drops only when we meet hibernation token. Then, this will work naturally for streamed input, and for the regular `.parse` you can just feed input in chunks (basically what it does now, but on the higher level).

inikulin: We will need to introduce special flag that will signal that we met hibernation token which we will need to check in initial states and then reset it. So we have a lot of duplicate code or we'll need to introduce additional method call (by refactoring flag check + preprocessor.dropParsedChunk + flag reset into separate method of Tokenizer). Also, with small chunks we may have opposite effect, since we'll produce a lot of intermediate strings. Overall, I'm not sure about this approach to be honest.

RReverser: Maybe you're right... Just trying to think of the cases where we will still have congestion inside of the streaming parser when chunks were already fed to it, but it didn't output the text node (because it was all text).

Is `bufferWaterline` supposed to be configurable via options btw?

inikulin: > Just trying to think of the cases where we will still have congestion inside of the streaming parser when chunks were already fed to it, but it didn't output the text node (because it was all text).

We will not have such issues with text content, but we will have such issue e.g. with extremely big attribute names/values. In that case our mechanism will not work and we'll buffer huge chunk of data.

> Is bufferWaterline supposed to be configurable via options btw?

I'm not sure if we should. It doesn't pass basic assertion for public API: how you will describe it to users? Maximum amount of data buffered by parser? It's not true, because we don't buffer exactly this amount and also we have edge cases there we can potentially buffer much more than that limit. I believe we should keep it private, since I would like to leave a space for improvement maybe we'll drastically modify this mechanism to handle edge cases.

package.json
inikulin: I've removed node .10 from test matrix and switched to native Promise

lib/tokenizer/preprocessor.js
RReverser: Ok, we can revisit this in future should someone need it.

PR #155, opened 2016-09-30
fix(serializer): safe use of replace

teleaziz:
when server-rendering angular2 apps using angular/universal I get str.replace is not a function error, debugging lead me here

### DIFF
--- a/lib/serializer/index.js
+++ b/lib/serializer/index.js
@@ -37,7 +37,7 @@ var Serializer = module.exports = function (node, options) {
 
 // NOTE: exported as static method for the testing purposes
 Serializer.escapeString = function (str, attrMode) {
-    str = str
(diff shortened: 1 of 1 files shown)


domenic:
It would be more interesting to find the bug in the code that was passing this a non-string. Parse5 seems correct.

gdi2290:
Parse5 is correct you don't need to merge. The code for Universal is not ensuring the property is a string when setting the attribute. The bug showed up when we started to convert DOM properties to attributes since users weren't realizing that they should be setting attributes for DOM APIs.

see fix https://github.com/angular/universal/pull/567 in Angular Universal

teleaziz:
glad you guys figured it out at universal, I still think type checking is necessary here as well.

domenic:
There is type checking! It throws a TypeError if you pass the wrong type :)

PR #159, opened 2016-10-19
Fixed incorrect LocationInfo.endOffset for non-implicitly closed elements

wooorm:
3a3515a seems to have introduced an error for all other elements (non-implicitly closed elements), as can be seen in the included test.

Kinda weird that it wasn’t caught before!

Cheers


inikulin:
Nice catch! Thank you for the contribution.

inikulin:
Released as `v2.2.3`

wooorm:
Perfect, thanks! 🎉

PR #161, opened 2016-10-25
Make sure attrs is defined on createElement()

pik:
* fixes parse5.serialize() breaking on elements created via. the default treeAdapter

### DIFF
--- a/lib/tree_adapters/default.js
+++ b/lib/tree_adapters/default.js
@@ -58,6 +58,7 @@ exports.createDocumentFragment = function () {
  *
  */
 exports.createElement = function (tagName, namespaceURI, attrs) {
+    if (!attrs) attrs = []
     return {
         nodeName: tagName,
         tagName: tagName,


inikulin:
Hi, thank you for contribution. Can you elaborate, please, what problem this PR solves.

pik:
Yes, currently elements returned by `createElement()` are not serialize-able e.g. the following will fail with:  `cannot read property length of undefined`

```es6
const adapter = parse5.treeAdapters.default
const document = adapter.createDocument()
const element = adapter.createElement('div')
adapter.appendChild(document, element)
parse5.serialize(document)
```

This is because serialize expects any initialized node to have attrs set to `[]`. Compare to the `parse` method which returns a usable element right away:
```es6
> p.parse('<div></div>')
{ nodeName: '#document',
  quirksMode: true,
  childNodes: 
   [ { nodeName: 'html',
       tagName: 'html',
       attrs: [],
       namespaceURI: 'http://www.w3.org/1999/xhtml',
       childNodes: [Object],
       parentNode: [Circular] } ] }
```

inikulin:
Well, it's something that shouldn't be fixed on parse5 side. `attrs` argument [is marked as required in API reference](https://github.com/inikulin/parse5/wiki/Documentation#TreeAdapter.createElement). And parser always passes this attribute. `treeAdapter` is not a general purpose utility library for manipulating AST, it's designed to maintain minimal API required by parser to build AST, anything else is out focus of this project. Any additional functionality on top of parse5 AST can be found in following utility libraries:
- https://github.com/webdeps/parse5-utils
- https://github.com/PolymerLabs/dom5

pik:
Sure it's required in the specification but that doesn't make it a good choice.

I'm really not sure why you'd rather point people to using a wrapper library which basically no-one is using, rather than adding one line to the existing code (one line which has absolutely no drawbacks!).

inikulin:
> rather than adding one line to the existing code (one line which has absolutely no drawbacks!)

It has a drawback that it reduces performance by introducing additional check that is not required by parser because it always provides `attrs` argument. Also, it's a question of consistency, what makes `attrs` argument special that it can be omitted? Should we do it for other arguments and functions? And once again: **`treeAdapater` is not a general purpose AST manipulation library, it exists just to tell the parser how to deal with current tree format**. 

> I'm really not sure why you'd rather point people to using a wrapper library which basically no-one is using

I'm just trying to help you solve your problem.

PR #169, opened 2016-12-05
fixed name and link authors

webdesus:
fixed name and link authors


inikulin:
Haha, good catch. Can you regenerate docs as well by running `gulp docs`?

webdesus:
I did it 😊

PR #186, opened 2017-03-17
Change extensions machinery

inikulin:


.eslintrc
```diff
@@ -30,7 +30,7 @@
     "no-shadow": 2,
     "no-shadow-restricted-names": 2,
     "no-spaced-func": 2,
-    "no-trailing-spaces": 2,
+    "no-trailing-spaces": 0,
```
RReverser: Why this change? Trailing spaces aren't nice :)

inikulin: I've got tired of this thing being buggy in some cases and finally disabled it in all my projects. Yes, we can increase file size due to unnecessary invisible symbols, but we definitely will not break anything with it.

inikulin: On the other hand, maybe it's just me being too impulsive at the end of the hard day. I've not decided yet, tbh 😄

RReverser: It's not about file size, but rather about annoyance, especially when editor is configured to auto-trim them (e.g. via `.editorconfig`) and then you're getting weird diffs whenever you try to commit changes.

inikulin: For some reason mine ignores this `.editorconfig` rule completely 😢. Reverted: fabc6d8eaba7904ae667dd20741bed9ea1a39ca0

PR #196, opened 2017-05-16
Fix empty stream parsing with ParserStream

stevenvachon:
This needs a test, but don't know which file to put it in.

### DIFF
--- a/lib/parser/parser_stream.js
+++ b/lib/parser/parser_stream.js
@@ -35,7 +35,7 @@ ParserStream.prototype._write = function (chunk, encoding, callback) {
 
 ParserStream.prototype.end = function (chunk, encoding, callback) {
     this.lastChunkWritten = true;
-    WritableStream.prototype.end.call(this, chunk, encoding, callback);
+    WritableStream.prototype.end.call(this, chunk || '', encoding, callback);
 };
 
 //Scriptable parser implementation
--- a/test/fixtures/parser_test.js
+++ b/test/fixtures/parser_test.js
@@ -126,3 +126,12 @@ exports["Regression - Don't inherit from Object when creating collections (GH-11
     }
 };
 
+exports['Regression - Fix empty stream parsing with ParserStream (GH-196)'] = function (done) {
+    var parser = new parse5.ParserStream()
+        .once('finish', function () {
+            assert(parser.document.childNodes.length > 0);
+            done();
+        });
+
+    parser.end();
+};


RReverser:
Judging from the existing structure, I'd suggest to put one into [test/data/tree_construction_regression](https://github.com/inikulin/parse5/tree/master/test/data/tree_construction_regression) - it will be automatically tested both with regular and streaming mode by [test/fixtures/parser_test.js](https://github.com/inikulin/parse5/blob/e2f5b5872247265431f6a175ab7a290512eac2f5/test/fixtures/parser_test.js#L54-L55).

lib/parser/parser_stream.js
```diff
@@ -28,12 +29,16 @@ inherits(ParserStream, WritableStream);
...
 //WritableStream implementation
 ParserStream.prototype._write = function (chunk, encoding, callback) {
+    this.written = true;
     this.writeCallback = callback;
     this.parser.tokenizer.write(chunk.toString('utf8'), this.lastChunkWritten);
     this._runParsingLoop();
 };
 
 ParserStream.prototype.end = function (chunk, encoding, callback) {
+    if (!this.written && typeof chunk === 'undefined')
```
RReverser: Btw, `chunk === undefined` would be equivalent and simpler. Or you can even do `chunk = chunk || ''`.

inikulin: Or even better `chunk === void 0`

stevenvachon: The lint doesn't allow it.

inikulin: BTW, I believe it's always safe to write empty string here, so I guess the best way will be to eliminate `this.written` variable and use the second code snippet suggested by Ingvar.

stevenvachon:
The issue solved by this PR should've already been tested by https://github.com/inikulin/parse5/blob/master/test/data/tree_construction/tests1.dat#L839-L846

unless it wasn't ran by `ParserStream`.

inikulin:
@stevenvachon If this issue was covered by this test then why it didn't fail?

inikulin:
This test doesn't test that we return well formed document in that case.

stevenvachon:
@inikulin I'm asking you, since it's your library.

stevenvachon:
It seems to assert the error instead of the document. Is this a bug in the test utils?

inikulin:
We don't even need a test file in this particular case, since it's just an empty string and bug occurs in very specific scenario. Take a look at other regression tests for parser for a clue.

stevenvachon:
If we don't need a test, how are we to ensure that this fix is not undone in the future?

inikulin:
> @inikulin I'm asking you, since it's your library.

idk, it's you claiming that adding new tests is redundant. And I don't understand why, since we don't have tests that cover this particular case. And no, this test shouldn't caught this case, because if you take a look in the code of test utils we'll have a single chunk with `''` content which is written to the parser. Therefore, it doesn't cover our case.

> If we don't need a test, how are we to ensure that this fix is not undone in the future?

I've never claimed that we don't need a test, I wrote that we don't need a test file like this https://github.com/inikulin/parse5/blob/master/test/data/tree_construction/tests1.dat. Since error is caused by the specific combination of API and not by the particular input you'll need to write test in the same format as e.g. this tests: https://github.com/inikulin/parse5/blob/master/test/fixtures/parser_test.js#L62

stevenvachon:
The confusion was "this particular case", where "case" sounded like "this PR". And my questions were asked to save me time, since you're familiar with the code.

This PR should be ready to go now.

inikulin:
@stevenvachon seems like we've had a mutual misunderstanding. LGTM, thank you for this!

stevenvachon:
No problem. Cheers.

Issue #197, opened 2017-05-27
No location info for inline elements containing block elements


stevenvachon:
```js
const html = `<a href="#"><p></a>`;
const document = require("parse5").parse(html, { locationInfo:true });

document.childNodes[0].childNodes[1].childNodes[0].__location;
//-> { line:1, col:7, startOffset:6, endOffset:25, ... }

document.childNodes[0].childNodes[1].childNodes[1].childNodes[0].__location;
//-> undefined
```

inikulin:
Congratulations, you've hit the darkest corner in the HTML parsing that is the [Adoption Agency Algorithm](https://html.spec.whatwg.org/multipage/syntax.html#adoption-agency-algorithm). 
Consider the following markup `<a>X<p>Y</a>` (it's a bit modified version of your markup with text nodes added for clarity), here is how it will be parsed:
![image](https://cloud.githubusercontent.com/assets/453071/26523057/b31d3670-4316-11e7-987d-f33c05472a15.png)

Parser encounters `a` element that is "split" by `p` element. `a` can't contain `p` element since the former is a block element, therefore it "squeezes out" `p` out of the `a` element. However, according to the markup part of the `p` content was also in the `a` element content. Thus, to preserve semantics and styling parser creates an additional **fake** `a` element (with the same attributes as the original `a` element) inside `p` element that incapsulates content that was previously in both elements.

And here we finally come to our issue: the element whose location you try to obtain is that **fake** `a` element. It doesn't exists in the markup, hence it doesn't have any location. However, the original `a` element (i.e. `document.childNodes[0].childNodes[1].childNodes[0]`) has location as seen from your example.

To conclude it's a by design behavior: we don't add location for the nodes that are not present in the markup, which IMHO makes perfect sense.

stevenvachon:
Thank you. Perhaps we should mention this algorithm in the documentation as an example of when `__location` will be `undefined`. Also, the documentation describes its value as "null", which isn't strictly accurate.

inikulin:
@stevenvachon Would love to accept PR for this

PR #198, opened 2017-06-08
Improve location info documentation

stevenvachon:
As discussed https://github.com/inikulin/parse5/issues/197#issuecomment-304474031

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -54,9 +54,9 @@ declare namespace MarkupData {
 declare namespace Options {
     export interface ParserOptions {
         /**
-         * Enables source code location information for the nodes. When enabled, each node (except root node) has the `__location` property.
-         * In case the node is not an empty element, `__location` will be {@link MarkupData.ElementLocation} object, otherwise it's {@link MarkupData.Location}.
-         * If the element was implicitly created by the parser it's `__location` property will be null.
+         * Enables source code location information. When enabled, each node (except the root node) will have a `__location` property.
+         * If the node is not an empty element, `__location` will be a {@link MarkupData.ElementLocation} object, otherwise it will be {@link MarkupData.Location}.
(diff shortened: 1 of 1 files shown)


lib/index.d.ts
```diff
@@ -54,9 +54,9 @@ declare namespace MarkupData {
 declare namespace Options {
     export interface ParserOptions {
         /**
-         * Enables source code location information for the nodes. When enabled, each node (except root node) has the `__location` property.
-         * In case the node is not an empty element, `__location` will be {@link MarkupData.ElementLocation} object, otherwise it's {@link MarkupData.Location}.
-         * If the element was implicitly created by the parser it's `__location` property will be null.
+         * Enables source code location information. When enabled, each node (except the root node) will have a `__location` property.
+         * If the node is not an empty element, `__location` will be a {@link MarkupData.ElementLocation} object, otherwise it will be {@link MarkupData.Location}.
+         * If the element was implicitly created by the parser (as part of the [Adoption Agency Algorithm](https://html.spec.whatwg.org/multipage/syntax.html#adoption-agency-algorithm)), it's `__location` property will be `undefined`.
```
inikulin: The adoption agency algorithm is only one of many cases (though the trickiest one) when the parser creates elements implictly.

stevenvachon: If it's not too much to ask, could you list them all here? I'll try to summarize them tersely.

inikulin: It's a lot of cases. I believe we shouldn't list them all in documentation either. Just give an example (https://html.spec.whatwg.org/multipage/syntax.html#an-introduction-to-error-handling-and-strange-cases-in-the-parser has a good explanation for some such cases, you can point to it).

stevenvachon: Done.

inikulin:
Thank you!

PR #202, opened 2017-06-21
Move @types/node to devDependency

7Pass:
Please make @types/node an development dependency so that it won't be installed when parse5 is used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
     "publish-please": "^2.2.0",
     "through2": "^2.0.0",
     "typedoc": "^0.5.1",
-    "typescript": "^2.0.6"
+    "typescript": "^2.0.6",
+    "@types/node": "^6.0.46"
   },
   "keywords": [
     "html",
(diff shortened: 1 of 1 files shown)


inikulin:
Why should we do this?

7Pass:
@types/node is only used for development, and is not used when parse5 is used as a library. By making @types/node a devDependency:

1. @types/node is installed when parse5 is referneced, even though it doesn't need to. A slightly longer installation time for an additional package that is not really used.
2. For my case, my project is using cheerio, which uses parse5, for some task during webpack bundling for a front end application with TypeScript. With this new update, @types/node is installed, causing typing conflict since our application is using front-end typings, not node typings.

inikulin:
> @types/node is only used for development, and is not used when parse5 is used as a library

It is used in production code in `.d.ts` file.

7Pass:
You're correct, I did not think about since our project does not use parse5 directly. I'll close the pull request.

inikulin:
@7Pass Please take a look at #235

PR #221, opened 2017-11-08
Add preserveCase option.

sihorton:
I needed to be able to preserve the attribute case for a SAX application I am writing. This pull request implements that change by modifying the `toAsciiLowerChar` function to return the result of `toChar` if a preserveCase option is set.

To eliminate the `if (preserveCase) {` test we could have a function pointer that is swapped between toAsciiLowerChar function and toChar function so there is just one test on initialization of the module. I can do another pull request with that option if it is interesting.


inikulin:
Hi, thank you for the PR. Unfortunately, we don't accept any changes that violate HTML specification.

sihorton:
ok, I understand.

PR #246, opened 2018-04-25
Fixing broken link to README.md file

davsm:
Underscore should be a dash.
Affects the code and the comment that is written to the auto-generated file
(ran into the broken link while troubleshooting)


inikulin:
Thank you!

PR #258, opened 2018-06-10
parser-stream docs: parens required when destructuring parameter

bathos:
Fixes a typo in an example in the docs for parse5-parser-stream.


inikulin:
Thanks!

PR #267, opened 2018-08-09
Fix SAXParser text location

### LINKED ISSUE #266: Incorrect location info in SAXParser and RewritingStream
When concatenating consequent character tokens, streaming variants of the parser don't correctly merge location information.

I found related issue #153 where `endOffset` was fixed for SAXParser, but `endLine` and `endCol` are still invalid and correspond to the first token.

Test for the `RewritingStream` (not for `SAXParser` as initially I thought the problem is specific to the rewriter) as an example of the issue:

```js
exports['RewritingStream - Should return location for entire text span'] = createRewriterTest({
    src: 'te\n st',
    expected: '',
    assignTokenHandlers: rewriter => {
        rewriter.on('text', token => {
            assert.deepStrictEqual(token, {
                text: 'te\n st',
                sourceCodeLocation: {
                    startLine: 1,
                    startCol: 1,
                    startOffset: 0,
                    endLine: 2, // returns 1 instead
                    endCol: 4, // returns 3 instead
                    endOffset: 6
                }
            });
        });
    }
});
```

RReverser:
Fixes #266.

### DIFF
--- a/packages/parse5-sax-parser/lib/index.js
+++ b/packages/parse5-sax-parser/lib/index.js
@@ -87,7 +87,12 @@ class SAXParser extends Transform {
                     if (this.pendingText === null) {
                         this.currentTokenLocation = token.location;
                     } else {
-                        this.currentTokenLocation.endOffset = token.location.endOffset;
+                        const { endLine, endCol, endOffset } = token.location;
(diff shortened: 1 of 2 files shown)


RReverser:
Will open another PR instead.

PR #270, opened 2018-08-10
Accept only decoded string streams

### LINKED ISSUE #269: SAXParser assumes that chunks contain valid UTF-8
Currently SAXParser gets binary blobs from the input stream and converts each chunk individually into string. This won't work correctly where UTF-8 codepoint is split across chunk bounds.

RReverser:
Previously, all streams accepted buffer chunks, decoding each one individually into a string as UTF-8.

This is problematic due to several reasons:
 1. It's not guaranteed that a binary stream is indeed UTF-8, and parse5 doesn't currently implement encoding detection to properly recognise anything else.
 2. For any multibyte encoding, including UTF-8, it's not guaranteed that a binary chunk on its own is a valid text as a codepoint can be split between different chunks (this was not caught by tests because they already feed string chunks into the parser).
 3. Writable parts of the stream didn't have `decodeString: false`, which means that by default Node.js would unnecessarily "decode" (d'oh, that naming is confusing) any string into a Buffer, feed it into a parser stream, and it would decode that Buffer back to a string. This is wasteful.

This commit fixes these issues by adding `decodeString: false` to Writable constructors so that string are passed into `_write` and `_transform` implementations as-is, and adds assertions to make sure that input is indeed a string.

Technically this is a breaking change as now all consumers are expected to feed readable stream with correct encoding (this is easy to do by passing `encoding: "utf8"` to any standard readable stream), and not just binary chunks, so parse5 will start throwing for existing binary usages, but I believe this is better than silently accepting and potentially breaking raw input.

Another option to accept raw input would be to use `StringDecoder` internally, but this would bring us back to issue (1) in the list, as we would need to properly detect input encoding rather than just assume it's UTF-8, so I'd rather leave this as a future enhancement.

Fixes #269.

### DIFF
--- a/bench/memory/sax-parser.js
+++ b/bench/memory/sax-parser.js
@@ -37,7 +37,7 @@ async function main() {
 }
 
 async function parse() {
-    const data = fs.readFileSync('../test/data/huge-page/huge-page.html');
+    const data = fs.readFileSync('../test/data/huge-page/huge-page.html', 'utf8');
     let parsedDataSize = 0;
     const stream = new SAXParser();
 
--- a/bench/perf/index.js
+++ b/bench/perf/index.js
@@ -107,7 +107,7 @@ runBench({
         const parsePromises = files.map(
             fileName =>
                 new Promise(resolve => {
-                    const stream = createReadStream(fileName);
+                    const stream = createReadStream(fileName, 'utf8');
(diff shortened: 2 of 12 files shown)


RReverser:
@inikulin Rebased and fixed conflicts.

PR #279, opened 2018-10-09
allow serialize callers to provide optional string escape callback

stevvvn:
I ran into [a situation](https://github.com/PolymerX/lit-loader/pull/5) where this library was used to reassemble a document, but it was a document including some template syntax that needs not to be converted to HTML entities (`>` => `&gt;`).

Granted the result is not valid HTML5 and thus arguably out of scope here, but the option to provide a custom escaping strategy might have some further use to customize handling of higher UTF chars or something.

### DIFF
--- a/packages/parse5/docs/options/serializer-options.md
+++ b/packages/parse5/docs/options/serializer-options.md
@@ -8,6 +8,17 @@
 
 ## Properties
 
+<a id="escapeString"></a>
+
+### `<Optional>` escapeString
+
+**● escapeString**: *function(str, attrMode)*
+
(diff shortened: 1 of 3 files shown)


inikulin:
Hello, thank you for the efforts, but unfortunately I can't accept this PR as any others issued previously that try to introduce any kinds of hooks that will allow people to use parse5 for what it hasn't been intended to be used (HTML parsing in serialization).

Though, as a workaround, I can suggest to take a look at [parse5 rewriting stream](https://github.com/inikulin/parse5/blob/master/packages/parse5-html-rewriting-stream/docs/index.md) which may give more fine grained control over the output for this particular case.

PR #283, opened 2018-11-12
Fix docs for Location info: endOffset & endCol

meisl:
In the files under https://github.com/inikulin/parse5/tree/master/packages/parse5/docs/source-code-location it says:
* `endOffset`: Zero-based index of the last character
* `endCol`: One-based column index of the last character

That's incorrect: they do not point *at* the last character but *after* the last character (which is as it should be, like String.substring or String.slice).

Note that `endLine` is described correctly as "One-based line index of the last character".
That is, this (one-based) index really points *at* the line that contains the last character.
Again, that's how it should be. However, maybe add a note to this effect - like the above - to the docs?

For now, I've just added "plus 1" and ", ie. it points *after* the last character" for `endOffset` and `endCol`.

I'll happily amend this PR in case of any suggestion for better and/or additional formulations.

### DIFF
--- a/packages/parse5/docs/source-code-location/element-location.md
+++ b/packages/parse5/docs/source-code-location/element-location.md
@@ -31,7 +31,7 @@ ___
 
 **● endCol**: *`number`*
 
-One-based column index of the last character
+One-based column index plus 1 of the last character, ie. it points *after* the last character
 
 ___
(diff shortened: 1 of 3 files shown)


fb55:
Thanks for flagging @meisl! Included a fix for this in #283.

PR #305, opened 2019-11-05
Fix document type detection (limited-quirks mode)

squidfunk:
This PR fixes a bug with `DOCTYPE` detection, where the lowercased `publicId` will never match the `limited-quirks` document types with mixed case.


inikulin:
@squidfunk Good catch, thanks! Though, seems like linting in CI is unhappy with line endings.

squidfunk:
Fixed, sorry for that.

squidfunk:
BTW, I don't know if the spec changed since you implemented quirks-mode detection, but I think the prefixes are not entirely correct as they specify `//en` at the end. See [this section](https://html.spec.whatwg.org/multipage/parsing.html#the-initial-insertion-mode:quirks-mode) of the spec which doesn't specify them. Also my testing with `document.compatMode` denotes that the language must not necessarily be `en` for correct quirks-mode detection.

inikulin:
@squidfunk Thanks for yet another great catch. I've just looked through the spec commit history and it seems like it always was this way and it's a mistake on my side. I'm surprised it wasn't captured by html5lib-tests.

May I ask you to fix that in context of this PR as well and open a ticket in https://github.com/html5lib/html5lib-tests to add test cases for that?

squidfunk:
> May I ask you to fix that in context of this PR as well and open a ticket in https://github.com/html5lib/html5lib-tests to add test cases for that?

Will do.

squidfunk:
I could combine some of the prefixes and add two missing - should now be spec-compliant.

squidfunk:
It would be really awesome if you could issue a bugfix release, as my downstream project pretty much depends on this fix :-)

inikulin:
@squidfunk Great work, thank you!

> It would be really awesome if you could issue a bugfix release, as my downstream project pretty much depends on this fix :-)

Will do =)

inikulin:
@squidfunk Published as v5.1.1

squidfunk:
Perfect, thanks! Happy to contribute.

PR #308, opened 2020-01-17
ES5 Upgrades for better compilation to ES2020

frank-dspeed:


frank-dspeed:
ah anyway forget it i fork of and rename complet i look what methods i can keep and recode the rest as its only a html parser i will get it done on this weekend.

inikulin:
Hi, I'm reading through changes and a bit confused. I've expected it to be changes to `.mjs` syntax, but it doesn't seem to be a case. Also, requirement to transpile it for older versions of node concerns me. Can you provide more details about the motivation for this change, please?

frank-dspeed:
I want that it is more easy to maintain and deploy reuse in bundlers. we can export a CJS version i don't care but the main code should be ESNext. so that i can use it in node 13 high performant and easy to bundle as dependency also i want to be able to do composition more easy.

frank-dspeed:
the current changes are only a little part to make it more easy convert able but i will rewrite it complet.

inikulin:
As far as I am aware all bundlers support CJS. By "high performance" you mean tree shaking capabilities of some bundlers? I expect that in Node there will be no difference in performance between module and CJS version, or I'm missing something?

frank-dspeed:
@inkulin your right we can close this i have already complet forked from this. i will look what i keep and what i join into my simple dom implamentation i have now 3 parsers i will benchmark then.

if you want to see the new Next gerneration version where the code is read and useable by 3th you can watch it under @as-esm/parse5 where i need to put anyway more then enought old school modules so that they can get used again.

Issue #295, opened 2019-05-11
link to the API docs in the README.md?


Pomax:
I don't see a link in the README.md that lets people get to the docs that cover the API for this package... can that be added?

rattrayalex-stripe:
It looks like the api docs for the main package are here: https://github.com/inikulin/parse5/blob/master/packages/parse5/docs/index.md

other packages here: https://github.com/inikulin/parse5/blob/master/docs/list-of-packages.md

Pomax:
It does, but I'm far less interested in "finding those link" as I am making sure _everyone_ can find them because they're listed in the main readme.

Pomax:
I've filed https://github.com/inikulin/parse5/pull/311, we'll see if @inikulin still looks at this repo.

PR #311, opened 2020-02-29
Make the readme.md actual markdown

Pomax:
Also, added a link directly to the API docs because those were missing for some reason.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,37 +1,16 @@
-<p align="center">
-    <a href="https://github.com/inikulin/parse5">
-        <img src="https://raw.github.com/inikulin/parse5/master/media/logo.png" alt="parse5" />
-    </a>
-</p>
+[![](https://raw.github.com/inikulin/parse5/master/media/logo.png)](https://github.com/inikulin/parse5)
 
-<p align="center">
-<i>HTML parsing/serialization toolset for Node.js. <a href="https://html.spec.whatwg.org/multipage/">WHATWG HTML Living Standard (aka HTML5)</a>-compliant.</i>
-</p>
+HTML parsing/serialization toolset for Node.js. [WHATWG HTML Living Standard (aka HTML5)](https://html.spec.whatwg.org/multipage/)-compliant.
 
-<p align="center">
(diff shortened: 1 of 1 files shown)


inikulin:
Hi, thanks for the contribution, but it was not intentionally not an "actual" markdown, because the "actual" markdown now looks like this:
image

inikulin:
I would accept a PR that adds the link to documentation, though. But keep in mind that this is a monorepo that contains multiple packages, and `parse5` is just one of them.

Pomax:
I'm sorry, I don't see the problem? Is the issue that you need every centered? Because that seems... irrelevant for a decent readme? It's your call of course but then maybe call it readme.html since it's just an html file instead of a normal markdown file.

inikulin:
What I'm trying to say is that I don't understand what this change intended to improve. I see regression in how markdown is displayed. Also, I don't understand reasoning behind: the PR removes HTML elements from the README justifying it by the usage of the "actual" markdown. Though, it's worth mentioning that HTML is a subset of markdown. 

> Anyway, the link is there, should be a few seconds work for you to add it in the original HTML form.

True, but I didn't open this PR.

Pomax:
mostly it turned "not an md file" into an md file, but if you want to keep it html without renaming, that's your call. As for "True, but I didn't open this PR.", equally true but that keeps https://github.com/inikulin/parse5/issues/295 open of course, which you can probably resolve in a few seconds since you own the code and can just add in the link in the format you want instead directly without even needing a PR.

inikulin:
> probably resolve in a few seconds since you own the code and can just add in the link in the format you want, without even needing a PR, using the github "edit file" option.

This indeed shouldn't take long. Though, you probably don't realise that parse5 is just one of my projects. I have 5 projects that I'm actively working at the moment and a few dozens that require my attention from time to time. With that in mind, it's sometimes quite hard for me to even find these few minutes to not harm my work-life balance. 

After all, it's a FOSS project, so anyone can contribute this change if it really bothers them.

Pomax:
I have dozens of projects I'm part of as well, so I know the pain of wanting to work on something but literally never getting to it. I'll see if I can find some time to file a new PR instead.

PR #312, opened 2020-03-05
Fix null value passed to escapeString caused error

xiangdewei:
Issue:
If null value passed to escapeString  function then replace function would fail.

Solution:
Fix this so that framework doesn't break.


RReverser:
Duplicate of https://github.com/inikulin/parse5/pull/145. You need to fix the underlying issue in the caller that is passing invalid AST.

PR #321, opened 2020-07-21
set hr tags as self closing

43081j:
as far as i can see, this is a typo.

it is currently causing failures in our linter (which depends on parse5) when parsing HR tags.

your tests are a little non-self-explanatory so if there's somewhere i can add one, please do let me know.

### DIFF
--- a/packages/parse5/lib/parser/index.js
+++ b/packages/parse5/lib/parser/index.js
@@ -1546,7 +1546,7 @@ function hrStartTagInBody(p, token) {
 
     p._appendElement(token, NS.HTML);
     p.framesetOk = false;
-    p.ackSelfClosing = true;
+    token.ackSelfClosing = true;
 }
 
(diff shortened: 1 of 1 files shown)


inikulin:
Nice catch, thanks. Tests need to be added in https://github.com/html5lib/html5lib-tests/tree/master/tree-construction (html5lib-tests is a default HTML test suite, used by all spec implementers)

43081j:
there's plenty of test cases in there which use a `<hr>` so im not sure how they didn't catch this.

not entirely sure what the test is that would cover this given that...

jorenbroekema:
Try `<hr />` maybe? Because that's the case where the linter errors where we caught it.

inikulin:
I'm going to merge it now, as it was clearly a typo in the original code. However, it would be still great if you contribute a test for that.

inikulin:
Published as 6.0.1

PR #326, opened 2020-10-01
Make << in comment text parse correctly

### LINKED ISSUE #325: Bug: Inside comments, `<<` is parsed as `<!`
**Module**: parse5-sax-parser@6.0.1

**Repro steps** (Linux):

```sh
$ cd $(mktemp --directory)
$ npm i parse5-sax-parser
[... npm output installing parse5-sax-parser@6.0.1 ...]
$ node << 'EOF'
const Parser = require('parse5-sax-parser')
const p = new Parser()
p.on('comment', (c) => console.log(c))
p.end('<!--test <<-->')
EOF
{ text: 'test <!', sourceCodeLocation: undefined }
```

**Expected**: `test <<`, not `test <!`.

**Rationale**:  I find the above behaviour confusing because the [HTML spec on comments](https://www.w3.org/TR/2011/WD-html5-20110525/syntax.html#comments) does not limit how `<` can be used inside comments.  A comment containing 2 consecutive less-than signs should be legal, but is currently unrepresentable.

**Analysis**:

I've just walked into the source and don't know the details, but it appears wrong to me that the tokeniser [switches state from `COMMENT_STATE` to `COMMENT_LESS_THAN_SIGN_STATE` when encountering `<`](https://github.com/inikulin/parse5/blob/37227a3429584903cbd1799dade995266fc2dbe6/packages/parse5/lib/tokenizer/index.js#L1451).  `COMMENT_LESS_THAN_SIGN_STATE` then [treats `<` as `!`](https://github.com/inikulin/parse5/blob/37227a3429584903cbd1799dade995266fc2dbe6/packages/parse5/lib/tokenizer/index.js#L1470-L1471) and causes the weird output seen above.

Surely `COMMENT_STATE` represents the state where we're inside the text part of the comment?  The only non-error references out of there should be to itself (parsing more content) or to `COMMENT_END_DASH_STATE`, right?

anko:
Inside comments two consecutive less-than characters (`<<`) parsed wrongly as `<!`, due to what was probably a typo.  This fixes that.

Also added regression test.

Fixes #325.

### DIFF
--- a/packages/parse5/lib/tokenizer/index.js
+++ b/packages/parse5/lib/tokenizer/index.js
@@ -1468,7 +1468,7 @@ class Tokenizer {
             this.currentToken.data += '!';
             this.state = COMMENT_LESS_THAN_SIGN_BANG_STATE;
         } else if (cp === $.LESS_THAN_SIGN) {
-            this.currentToken.data += '!';
+            this.currentToken.data += '<';
         } else {
             this._reconsumeInState(COMMENT_STATE);
         }
--- a/packages/parse5/test/tokenizer.test.js
+++ b/packages/parse5/test/tokenizer.test.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const assert = require('assert');
+const parse5 = require('../lib');
 const path = require('path');
 const Tokenizer = require('../lib/tokenizer');
 const Mixin = require('../lib/utils/mixin');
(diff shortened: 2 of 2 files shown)


inikulin (approved):

inikulin:
@anko thank you!

anko:
Could a patch version be published with this fix?

I may be the only person on the planet routinely parsing shell scripts containing `<<` out of HTML comments, but it makes me happy. :sweat_smile:

PR #346, opened 2021-06-07
Do not escape ampersand (`&`) in href attribute

thecp:
Prior to this commit an ampersand would be escaped although it is
totally valid. This commit fixes that.


fb55:
Thanks for this patch! This unfortunately breaks serialisation of eg. `<a href="https://example.com/foo?bar=baz&ampampa=1">`.

Escaping all ampersands in URLs isn't ideal, but unfortunately this will need an a bit more measured approach.

PR #347, opened 2021-06-21
Add outerHTML option to serializer. Resolves #230.

### LINKED ISSUE #230: Add a serialization mode that mimics outerHTML
That is, it should include the node itself.

Currently in jsdom we have to create a fake parent node with only one child node, the node we want to serialize, and then ask parse5 to serialize that.

joeldenning:
See #230. I didn't add tests yet to this because I wasn't sure the best place to do so and also wasn't sure if this change would be welcomed/accepted. I'm happy to add tests after it's reviewed if things are looking good!

### DIFF
--- a/packages/parse5/lib/serializer/index.js
+++ b/packages/parse5/lib/serializer/index.js
@@ -11,7 +11,8 @@ const NS = HTML.NAMESPACES;
 
 //Default serializer options
 const DEFAULT_OPTIONS = {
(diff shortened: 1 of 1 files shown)


fb55:
Hi @joeldenning, thanks for this PR, agreed that this is a needed feature. I don't think an option is the way to go here; I would opt for a separate method instead.

Closing this PR for now. We will continue having #230 as a reference.

Issue #349, opened 2021-07-01
PR: backports parse6 bom_fix for cheerio issue


frank-dspeed:

43081j:
"parse6"? have you made a fork? have you opened wider discussion in an issue anywhere for that?

would be worthwhile understanding what lead to you making a fork and why you couldn't contribute it to the official package.

i ask because i've also got a branch where i've moved parse5 to ESM, with the aim of moving it to typescript after that. but HTML parsing is a messy job, so if you're already working on something please do open it for wider discussion.

(sorry to anyone else for hijacking the issue)

frank-dspeed:
@43081j https://github.com/inikulin/parse5/issues/330

frank-dspeed:
i am not sure if i know your ESM port alreadyi saw some but my rewrite is far more deep and i break a lot of the existing API.

because i do want to have it without "this" usage and a lot of more to also run it in production.

I already have full typescript support but i am not using .ts extension.

While my rewrite is really far and did correct a lot of internal messy stuff like the state tracking it is to early to publish it as publishing would force me  to keep the existing api.

And as sayed i break the whole api and did also rewrite the test harness. 

as i started with the refactoring i found all kind of small logic failures that simply worked because of lucky conditions.

my main goal is to get the code readable for some one who never saw it before is maybe the most importent step.

frank-dspeed:
@43081j oh and i want to tell you why i do not directly contribute here i did want to do so but i feel bad this has millions of users i can never break the api of this repo. so it was logical that a fork is needed

fb55:
As discussed in #350, it unfortunately doesn't seem like this is the right way to go about things.

PR #350, opened 2021-07-01
Closes: #349 - Fix issues when BOM is before TAG

frank-dspeed:
Added Test and all other needed info the issue has the rest of the information
Closes: #349

### DIFF
--- a/packages/parse5/lib/common/unicode.js
+++ b/packages/parse5/lib/common/unicode.js
@@ -71,7 +71,8 @@ exports.CODE_POINTS = {
     LATIN_SMALL_F: 0x66,
     LATIN_SMALL_X: 0x78,
     LATIN_SMALL_Z: 0x7a,
-    REPLACEMENT_CHARACTER: 0xfffd
+    REPLACEMENT_CHARACTER: 0xfffd,
+    BOM: 0xfeff
 };
 
 exports.CODE_POINT_SEQUENCES = {
--- a/packages/parse5/lib/tokenizer/index.js
+++ b/packages/parse5/lib/tokenizer/index.js
@@ -134,7 +134,7 @@ const NUMERIC_CHARACTER_REFERENCE_END_STATE = 'NUMERIC_CHARACTER_REFERENCE_END_S
 //this functions if they will be situated in another module due to context switch.
 //Always perform inlining check before modifying this functions ('node --trace-inlining').
 function isWhitespace(cp) {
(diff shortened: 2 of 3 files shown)


frank-dspeed:
Closes: #349

fb55:
This doesn't seem to be in line with the spec: https://html.spec.whatwg.org/#event-stream-interpretation

> The UTF-8 decode algorithm strips one leading UTF-8 Byte Order Mark (BOM), if any.

It sounds like this should be done when decoding the original byte stream. This is implemented eg. in https://github.com/jsdom/whatwg-encoding

frank-dspeed:
@fb55 what changes would you propose ? while keeping it working?

fb55:
I would point users at whatwg-encoding if they are parsing content that isn't properly decoded yet. BOMs are a concern of a different part of the stack and out of scope for this library imho.

I am now thinking about adding a method to `cheerio` that allows users to load a buffer, which includes the `whatwg-encoding` module. I've opened https://github.com/cheeriojs/cheerio/issues/2051 to track this.

frank-dspeed:
@fb55 i agree that was my conclusion also over the long run to put that into a file loading layer! as that is a problem with files not with strings as strings normaly do not contain boms normaly.

but preserving and handling the BOM as whitespace does not hurt and preserves the position data which is also a strong coupled part at present

i still have a hard time to decouple the position counting and that is file based.

when we think about scenarios where we have a input file that contains a bom then the bom can not be save stripped. without messing up the position data "sourceCodeLocationInfo"

frank-dspeed:
# So Short Story at point of time this is the right thing to do out of my view

frank-dspeed:
@fb55 i would love if you could give me some feedback about my thinking process as i value your work.

fb55:
I think this is where we have slightly different ideas of what this module should be. To me, `parse5` is the module that closely follows the HTML spec and doesn't show any behaviour that isn't part of the spec.

> over the long run to put that into a file loading layer

That file loading layer already exists, with https://github.com/jsdom/whatwg-encoding and https://github.com/jsdom/html-encoding-sniffer. These follow the [WHATWG Encoding Standard](https://encoding.spec.whatwg.org/), which further guarantees compatibility with browsers. JSDom is using it, and it seems easy enough to add it to eg. Cheerio as well.

frank-dspeed:
@fb55 ok so i will revert that and add a option to Handle BOM as WhiteSpace and without that it get stripped with warning or error? 

What would be the correct behavior stripping it complet or not handle it at all? Should it get Replaced?

Issue #333, opened 2021-03-07
Serialization special casing of style, script, etc. needs to be namespace-dependent


domenic:
This escaping: https://github.com/inikulin/parse5/blob/9c683e154b8bbdc94237580680a934348406d1a2/packages/parse5/lib/serializer/index.js#L136-L149 is meant to mimic [HTML's](https://html.spec.whatwg.org/#serialising-html-fragments)

> If the parent of current node is a style, script, xmp, iframe, noembed, noframes, or plaintext element, or if the parent of current node is a noscript element and scripting is enabled for the node, then append the value of current node's data IDL attribute literally.

However, while HTML is referring specifically to those nodes in the HTML namespace, parse5 does not check the namespace before serializing.

This leads to the following wrong result in jsdom: https://runkit.com/domenicdenicola/6044168f463fb4001a725307 compared to browsers: http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8981

domenic:
https://github.com/inikulin/parse5/blob/9c683e154b8bbdc94237580680a934348406d1a2/packages/parse5/lib/serializer/index.js#L70-L89 is similarly suspicious... in general any checks against `$.SOMECONSTANT` really should be accompanied with a namespace check.

nfriedly:
This just bit me today: in a `<script>` tag `&&` is getting changed to `&amp;&amp;`, which is a syntax error.

nfriedly:
Actually, looking closer, `$.SCRIPT` is on that list, so I'm not sure if I'm hitting this or some other issue...

nfriedly:
Yeah, this is what's causing my issue:

https://github.com/inikulin/parse5/blob/9c683e154b8bbdc94237580680a934348406d1a2/packages/parse5-html-rewriting-stream/lib/index.js#L80

It doesn't even try to check the namespace or parent element there...

I'll file a separate issue.

Issue #332, opened 2021-03-06
Make serialization scriptingEnabled-dependent


domenic:
https://html.spec.whatwg.org/#serialising-html-fragments says

> If the parent of current node is a style, script, xmp, iframe, noembed, noframes, or plaintext element, or if the parent of current node is a noscript element **and scripting is enabled for the node**, then append the value of current node's data IDL attribute literally.

parse5 does not seem to implement this additional condition, and doesn't seem to even have the ability to tell the serializer about scripting-enabled or not: https://github.com/inikulin/parse5/blob/9c683e154b8bbdc94237580680a934348406d1a2/packages/parse5/lib/serializer/index.js#L136-L149

This is causing [this web platform test](https://github.com/web-platform-tests/wpt/blob/master/html/syntax/serializing-html-fragments/escaping.html) to fail in jsdom.

Issue #330, opened 2021-03-02
PR: 7.0 - ESM - Only version 


frank-dspeed:
As NodeJS LTS Now Reached the state where ESM Works and we got a 100% Way to shim it for older NodeJS Versions it
is time to Refactor this to ESM Only which would directly run Inside Nodejs & WebBrowsers 

## Issus to Tackle in this Major
- https://github.com/inikulin/parse5/issues/332
- https://github.com/inikulin/parse5/issues/333
- https://github.com/inikulin/parse5/issues/334
- https://github.com/cheeriojs/cheerio/issues/1506
- https://github.com/inikulin/parse5/pull/327

## Old Issue that got noisy and out of topic discussions
- https://github.com/inikulin/parse5/issues/307

## TODO
- [ ] Update Documentation
- [x] Drop use of class, this, null
- [X] Switch to Functional Composition and drop Mixins
- [X] Get TypeScript Happy with JSDOC Comments.
- [x] Create new Tests that are Working in NodeJS and the Brwoser.
- [x] Current Extension Model Replacement
  - [ ] Static Analyzeable Classes that do clear extend each other for all 3 situations Raw/with error/with sourcecode info
  - [x] Functional Composition 
  - [x] Event Emitter Pattern

## 5th iteration conclusions
Findings:
The Result of new Parser() should get wrapped when extending the parser it is not so easy to follow the inharitance of objects with
the self written Mixin Class system that offers a later install method which will patch a eventual existing _bootstrap method that later 
gets eventual called when using Parser.parse or Parser.parseFragment.

All extensions should be written as proxy the treeAdapter should be named NodeAdapter is it is not forming the tree out of browser view
it is a Node api so we should express it like that to make more clear for other developers that there is the place to manipulate Node
Creation and handling. Proxy support starts with Nodejs 6.4 shim https://github.com/GoogleChrome/proxy-polyfill/blob/master/src/proxy.js

- [x] v8 JSONParse Optimization https://v8.dev/blog/cost-of-javascript-2019
- [x] treeAdapter should get replaced by a new Node() hireOrderFunction.
- [ ] ErrorProxyHandler.applyl(target,thisArg, args) where target is Parser,Tokenizer,OpenElementStack, Preprocessor).
- [ ] srcCodeLocationInfo.apply(...) where target is Tokenizer and Preprocessor
- [ ] debuggingProxy as new feature that logs propertys.
- [x] Object.create(null) replaced by {}
[...]

PR #351, opened 2021-07-11
Move entire codebase to ESM

43081j:
Here i've moved the whole codebase to be written as ES modules.

This would be a huge breaking change. I have dropped all commonjs support.

This also resulted in fixing several bugs and bad tests that went unnoticed with CJS.

The aim here is to do another separate change which also moves the whole repo to typescript.

Again, this is a massive breaking change, completely backwards-incompatible by choice. If you don't want it, that is perfectly fine, i can keep it in a fork. If there's any changes and you want to help get it back into this repo, though, im happy to discuss/change whatever.

FYI all tests pass in its current state.

Note also, this probably _will not_ work in browsers. I'd love a world where node can handle true ESM, but it can't. This relies on node-only interop so our dependencies work (as our dependencies still use commonjs). In order to have a true esm codebase, all dependencies would also need to migrate.

cc @inikulin


43081j:
im still not following, why would a bundler struggle with a named export? webpack, rollup, esbuild, etc handle named exports just fine afaik. does one of you have a concrete example of some situation that won't work with a named export?

Konakona-chan:
@43081j , thank you for your answer (yeah, I see the pull-request is kind of stale, unfortunately)!

Just a note: after changing [only one line](https://github.com/syntax-tree/hast-util-raw/blob/main/lib/index.js#L70) this pull-request works great (well, I haven't seen any issue yet) in svelte/vite/unifiedjs ESM-only project. Good luck with reviewers in this pull request or forking.

fb55:
The benchmark script still had some broken imports; see https://github.com/parse5/parse5-fork/commit/f8afdbb8210fe80814d6463e6beb23dfe8298ed4 for a fix.

43081j:
good catch, have pushed a fix

fb55:
@43081j I have created a fork at https://github.com/parse5/parse5-fork the other day, and have invited you to join it. This PR, plus some other small updates, are already on `master`.

I have created a discussion of what to do with the fork at https://github.com/parse5/parse5-fork/discussions/2 — let's discuss next steps there.

fb55:
Thanks for this @43081j!

domenic:
Oh, this is a shame. So jsdom cannot depend on parse5 anymore?

fb55:
Hi @domenic, just for context: Could you summarise the stance of jsdom with regards to ESM? I am generally okay with delaying ESM in parse5 if this is a blocker somewhere else.

domenic:
jsdom has many CommonJS-only dependencies, so it will be CommonJS until all of its dependencies update and the maintainers have time for a conversion. It'd be much preferable if this package supported CommonJS (e.g. via one of the techniques [here](https://2ality.com/2019/10/hybrid-npm-packages.html) until that time; it seems like the only way that we're going to be able to move the ecosystem is with such hybrid strategies.

fb55:
Makes sense. I've opened #379 to track this.

wooorm:
I can understand wanting to stick with CJS, but I don’t quite get this argument:

> so it will be CommonJS until all of its dependencies update

You want to wait for *all users of JSDOM* to use ESM or dual, before JSDOM can switch to ESM? 🤔 I find that a weird argument, because:
- the same argument could be used for any breaking change but JSDOM is at v19 already
- some dependents of open source are abandoned / will never receive updates, so it’s not possible to ever reach 100%?

---

What about these alternatives:

- JSDom can use ESM-only packages, with a dynamic import. It matches your [`engines`](https://github.com/jsdom/jsdom/blob/a604d67995fe5b2b1ce73090d347c23eeed011eb/package.json#L114). It does require asyncness though, unfortunately.
- JSDOM can ship a bundled CJS version of parse5, quickly generated with esbuild

43081j:
@domenic is right though, in that until JSDOM's dependencies are also ESM, jsdom isn't _technically_ ESM (it'd depend on node interop).

jsdom could deliver a CJS bundle, but its still a pretty big chunk of work for them to convert their codebase to ESM.

old unmaintained packages tend to get replaced soon enough, fortunately. but its true thats often a blocker for moving to ESM if there's no replacement.

we could temporarily ship a CJS bundle until consumers have had time to move, that isn't difficult (have two tsc configs or use a bundler to convert it back to commonjs).

domenic:
>  i can't see us introducing any extra features anyway so @domenic could probably pin to 6.x

Well, several of the issues targeted for 7.0 per #330 add new features requested by me for jsdom specifically. Namely #332, #333, and #230.

domenic:
> You want to wait for all users of JSDOM to use ESM or dual, before JSDOM can switch to ESM?

No. I need to wait for all *dependencies* of jsdom to use ESM or dual.

wooorm:
> I need to wait for all *dependencies* of jsdom to use ESM or dual

As parse5 is a dependency of jsdom, and you are waiting for dependencies to use ESM or dual, then why do you now not want parse5 to become esm?

> Because ESM packages cannot depend on non-ESM packages without making package startup async

Could you elaborate? I have experience with ESM packages in native Node.js that use CJS, dual, and ESM dependencies with sync import statements.
And CJS packages that can only use ESM dependencies with dynamic `import()` expressions.

> #332, #333, and #230.

What if those also land on the current (CJS) version? (And future security stuff of course)

43081j:
@domenic totally missed those issues, my bad.

tbh i think we should just create a CJS bundle at build time until everyone's moved over. we don't want to block each other's progression to ESM and its a very simple thing for us to add.

PR #353, opened 2021-07-22
fix(parser): Set `endTag` loc for mixed-case foreign elements

### LINKED ISSUE #352: Missing `endTag` source location information on SVG elements with uppercased tag name
## Description

Parse5 doesn't report `endTag` source location for all the SVG elements with a tag name containing an upper case even when the end tag is present. eg: `foreignObject`, `linearGradient, etc.

## Repro steps

```js
const html = '<svg><foreignObject></foreignObject></svg>';
const fragment = parse5.parseFragment(html, {
  sourceCodeLocationInfo: true
});

const foreignObject = fragment.children[0].children[0];
console.log(foreignObject.sourceCodeLocation.endTag);

// Actual: undefined
// Expected: { startLine: 1, startCol: 21, startOffset: 20, endLine: 1, endCol: 37, endOffset: 36 }
```

pmdartus:
Update `setEndLocation` logic to account for tag name with uppercase.
Fixes #352.


fb55:
Thanks a lot @pmdartus!

PR #354, opened 2021-07-30
Add slot tag

mattvague:


fb55:
Thanks for opening this @mattvague! We'll handle this in one action with other spec changes later on.

PR #359, opened 2021-11-21
Typescript conversion of parse5 package

43081j:
All other packages have been left as-is.

I am not expecting this PR to ever land, it is more here as reference (or something?) until we decide what to do about the path forward (potentially a fork).

@fb55 sorry it took me so long to respond to you in the other PR. this is what I was busy doing meanwhile. it'll need somewhere to live, so we should find a place to discuss this since you seem the most willing to maintain a fork.


fb55:
Hi @43081j, great effort! I have already completed a TS port at https://github.com/parse5/parse5-fork, plus a lot of other quality of life improvements & bug fixes. I've invited you to the org in the past, let me know if you're interested in using that as a starting point. Quite unfortunate that conversion efforts were duplicated here :/

43081j:
yeah very unfortunate. its in a single commit really, which was started a few months ago so i doubt there was any chance we would've crossed paths.

there's a fair amount of stronger types in my branch than in yours, i've seen. if anything, i'd suggest you read through mine. meanwhile i'll have a think about what should happen with this as i feel like both branches have their good parts.

fb55:
Closing this, as the TS conversion was completed in #362. Great effort @43081j, and thanks for helping with my PR!

PR #360, opened 2021-11-28
chore(ci): Switch to GitHub Actions CI

fb55:
Also introduces Dependabot.

Travis.org is no longer supported, and the choice would have been to either switch to their .com offering, or switch CI providers entirely. GitHub Actions is the first party solution here, so let's use that!

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+version: 2
+updates:
+    - package-ecosystem: npm
+      directory: '/'
+      schedule:
+          interval: daily
+      open-pull-requests-limit: 10
+      versioning-strategy: increase
+    - package-ecosystem: 'github-actions'
+      directory: '/'
+      schedule:
+          interval: daily
+    - package-ecosystem: gitsubmodule
+      directory: '/'
+      schedule:
+          interval: daily
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,33 @@
+name: 'CodeQL'
+
+on:
+    push:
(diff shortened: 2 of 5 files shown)


wooorm:
👍 on GH CI.
Have you thought about something else for deps? I find that dependabot introduces a lot of noice. I particularly know that renovate can both preserve semver ranges and batch everything on, say, Monday?

fb55:
> Have you thought about something else for deps? I find that dependabot introduces a lot of noice. I particularly know that renovate can both preserve semver ranges and batch everything on, say, Monday?

My issue with anything but Dependabot is that it would have to be activated by @inikulin, as the owner of this repo.

Part of this PR is also a workflow that enables auto-merging for Dependabot PRs (not including major version bumps), which reduces the overhead of Dependabot a lot; I have this enabled for all of my repos.

It is possible to configure Dependabot to only update once a week. This doesn't reduce the noise a lot imho. Auto-merging PRs allows me to ignore Dependabot. I will go through the PRs before cutting a release, but generally things will just stay up-to-date.

wooorm:
> Part of this PR is also a workflow that enables auto-merging for Dependabot PRs (not including major version bumps)

Everything here is using `^`.
a) why run dependabot on those
b) why automatically merge a PR updating say `^7.0.2` to `^7.0.3`? It doesn’t do anything, except make installs harder to resolve and `package-locks` and `yarn.locks` grow a bunch

> this doesn't reduce the noise a lot imho.

My problem is that if a repo is creating noise that I don’t need to handle (e.g., dependabots every day), that I become desensitised to it. I start ignoring the repo. And I start missing things that do need my attention.
I get a lot of notifications each day already.

PR #327, opened 2020-11-04
Use domhandler node interface in htmlparser2 adapter

TrySound:
Ref https://github.com/cheeriojs/cheerio/issues/1506

cc @fb55 @inikulin

Issue #211, opened 2017-08-17
Add code coverage metrics


inikulin:
html5-lib tests can benefit from this as well, since we can find code paths for which it doesn't have tests.

PR #362, opened 2021-12-12
Move fork back into main project

fb55:
Ivan thankfully added some contributors, and it's time to move my fork back into the main project. Not sure yet about the best approach, open to suggestions.

Things left to do:
- [x] Write changelog for this PR
- [x] Merge PRs this PR is based on (#327 and #351)
- [x] Rename packages back to their original names
- [x] Fix imports to point at compiled files, so we can publish

---

## Changelog:

- Ported to TypeScript
- Switched to npm workspaces, in favour of `lerna`
- Switched to Jest as the test runner
- Switched to the `entities` module for decoding entities
  - A version of parse5's decoding logic was adapted for `entities`. Adopting this dependency allows us to share the maintenance with the `entities` & `htmlparser2` modules.
- Moved the docs back to TSDoc comments
- Switched to the state machine pattern of `htmlparser2` for tokenizer
- No more mixins: Merged location & error mixins into the main classes
- Introduced tag IDs, to avoid branching over large numbers of strings.
- Introduced Maps and Sets where appropriate
- Switched the order of the formatted elements list
- Introduced array helpers in a lot of places
- Updated tests to no longer build objects of tests (used `describe`/`it` constructs instead)


wooorm:
It’s a bit hard to review (I also get that it would be too much work to do it in many separate PRs). Anything you want me to look at?

wooorm:
sounds good!

wooorm:
unsubscribing from this while you’re still working on it as it’s a bit noisy. Do ping me when it’s ready!

fb55:
@wooorm This should be good to go now!

fb55:
Also for merging — I am considering a _Squash and merge_, what do you think?

wooorm:
I’m in favor of always squashing. I think the conversation of how a PR came to be should not be in the commits, as the discussion can be seen on the PR page.

With examples, I strongly prefer https://github.com/nodejs/node/commits/master over https://github.com/github/opensource.guide/commits/main

.eslintrc.json
```diff
@@ -2,7 +2,7 @@
     "env": {
         "es2020": true,
         "node": true,
-        "mocha": true
+        "jest": true
```
wooorm, line 5: I would rather use a different test mechanism, such as https://github.com/lukeed/uvu or maybe https://github.com/davidmarkclements/brittle.
Jest has a ton of problems with ESM, injects stuff, and provides much more overhead that needed

bench/package.json
```diff
@@ -1,15 +1,16 @@
...
     "description": "parse5 regression benchmarks",
     "author": "Ivan Nikulin <ifaaan@gmail.com>",
     "license": "MIT",
     "dependencies": {
         "benchmark": "^2.1.4",
-        "human-format": "^0.7.0",
+        "human-format": "^0.11.0",
         "@airbnb/node-memwatch": "^2.0.0",
-        "parse5": "*",
+        "parse5": "npm:parse5",
```
wooorm, line 13: Why?

lerna.json
```diff
@@ -1,5 +1,5 @@
 {
```
wooorm, line 1: Is lerna still needed?

package.json
```diff
@@ -2,41 +2,69 @@
     "name": "parse5-build-scripts",
     "private": true,
     "type": "module",
+    "workspaces": [
+        "packages/*",
+        "bench"
+    ],
     "devDependencies": {
+        "@types/dedent": "^0.7.0",
+        "@types/jest": "^27.4.0",
```
wooorm, line 11: How about loosening versions, e.g. `^27.0.0`, if they aren’t needed. It produces less noise in package(-lock) diffs and gives npm more room to choose what to install leading to less duplicates

package.json
```diff
@@ -2,41 +2,69 @@
...
+        "typescript": "^4.5.4"
     },
     "scripts": {
-        "build": "lerna run build",
-        "format": "prettier --write \"{packages/*/{lib,test}/**/*.js,{bench,test,scripts}/**/*.js}\"",
-        "bootstrap": "lerna bootstrap --no-ci",
-        "lint": "eslint --ext js .",
-        "unit-tests": "mocha --ui tdd --reporter progress --timeout 20000 packages/**/test/*.test.js",
+        "build": "npm run build --workspaces --if-present",
+        "prettier": "prettier '**/*.{js,ts,md,json,yml}'",
```
wooorm, line 29:
```suggestion
        "prettier": "prettier '**/*.{js,ts,md,json,yml}' --loglevel warn",
```

This prevents it from listing all the files it touches (most tools such as eslint don’t print that either)

packages/parse5-html-rewriting-stream/docs/index.md
```diff
@@ -1,163 +1,5 @@
 # Class: RewritingStream
 
-Streaming [SAX](https://en.wikipedia.org/wiki/Simple_API_for_XML)-style HTML rewriter. A [transform stream](https://nodejs.org/api/stream.html#stream_class_stream_transform) (which means you can pipe _through_ it, see example).
```
wooorm, line 3: Why remove docs?

tsconfig.json
```diff
@@ -1,18 +1,26 @@
 {
     "compilerOptions": {
-        "target": "es2020",
-        "module": "es2020",
+        "target": "es2019",
```
wooorm: why?

tsconfig.json
```diff
@@ -1,18 +1,26 @@
...
+        "allowSyntheticDefaultImports": true,
+
+        "outDir": "build",
+        "baseUrl": ".",
+        "paths": {
+            "parse5/dist/*": ["packages/parse5/lib/*"],
+            "parse5-sax-parser/dist/*": ["packages/parse5-sax-parser/lib/*"],
+            "parse5-htmlparser2-tree-adapter/dist/*": ["packages/parse5-htmlparser2-tree-adapter/lib/*"],
+            "parse5-parser-stream/dist/*": ["packages/parse5-parser-stream/lib/*"]
+        }
```
wooorm: `skipLibCheck: true` maybe?

tsconfig.json
```diff
@@ -1,18 +1,26 @@
 {
     "compilerOptions": {
-        "target": "es2020",
-        "module": "es2020",
+        "target": "es2019",
+        "module": "esnext",
```
wooorm: also: why?

packages/parse5-parser-stream/docs/index.md
```diff
@@ -1,45 +1,5 @@
 # Class: ParserStream
 
-Streaming HTML parser with scripting support. A [writable stream](https://nodejs.org/api/stream.html#stream_class_stream_writable).
```
wooorm, line 3: Why?

packages/parse5-sax-parser/docs/index.md
```diff
@@ -1,40 +1,5 @@
 # Class: SAXParser
 
-Streaming [SAX](https://en.wikipedia.org/wiki/Simple_API_for_XML)-style HTML parser. A [transform stream](https://nodejs.org/api/stream.html#stream_class_stream_transform) (which means you can pipe _through_ it, see example).
```
wooorm, line 3: Why?

packages/parse5-test-utils/data/parser-feedback/entities01.test
```diff
@@ -60,6 +60,26 @@
                 ]
             ]
         },
+        {
```
wooorm, line 63: Could this one be added in a separate PR, to emphasize that something important was fixed?

packages/parse5-test-utils/data/parser-feedback/noscript01.test
```diff
@@ -0,0 +1,436 @@
+{
+    "tests": [
```
wooorm, line 2: also maybe add later?

packages/parse5-test-utils/data/parser-feedback/tests16.test
```diff
@@ -2009,6 +2009,83 @@
                 ]
             ]
         },
+        {
```
wooorm, line 2012: add later?

packages/parse5-test-utils/data/parser-feedback/tests25.test
```diff
@@ -378,6 +378,153 @@
                 ]
             ]
         },
+        {
```
wooorm, line 381: later

packages/parse5-test-utils/data/parser-feedback/tests5.test
```diff
@@ -382,6 +382,33 @@
                 ]
             ]
         },
+        {
```
wooorm, line 385: later

packages/parse5-test-utils/data/parser-feedback/tests6.test
```diff
@@ -134,8 +134,8 @@
             ]
         },
         {
```
wooorm, line 136: Later

wooorm: Also: why is `\r` no longer supported?!

packages/parse5-test-utils/data/parser-feedback/tests7.test
```diff
@@ -427,8 +427,8 @@
             ]
         },
         {
-            "description": "<!doctype html><listing>\\r\\nX</listing>",
-            "input": "<!doctype html><listing>\r\nX</listing>",
+            "description": "<!doctype html><listing>\\nX</listing>",
```
wooorm, line 430: same as previous

packages/parse5-test-utils/data/parser-feedback/tests8.test
```diff
@@ -1,8 +1,8 @@
 {
     "tests": [
         {
-            "description": "<div>\\r\\n<div></div>\\r\\n</span>x",
-            "input": "<div>\r\n<div></div>\r\n</span>x",
+            "description": "<div>\\n<div></div>\\n</span>x",
```
wooorm, line 4: same as previous

packages/parse5-test-utils/data/parser-feedback/webkit02.test
```diff
@@ -56,6 +56,49 @@
                 ]
             ]
         },
+        {
```
wooorm, line 59: add in another PR

packages/parse5/package.json
```diff
@@ -6,6 +6,12 @@
     "author": "Ivan Nikulin <ifaaan@gmail.com> (https://github.com/inikulin)",
     "contributors": "https://github.com/inikulin/parse5/graphs/contributors",
     "homepage": "https://github.com/inikulin/parse5",
+    "scripts": {
+        "build": "rm -r dist; tsc"
+    },
+    "dependencies": {
+        "entities": "^3.0.1"
```
wooorm:
```suggestion
        "entities": "^3.0.0"
```

packages/parse5/docs/index.md
```diff
@@ -1,111 +0,0 @@
-# parse5
```
wooorm, line 1: Should all this information be moved to the readme?

packages/parse5-test-utils/package.json
```diff
@@ -0,0 +1,9 @@
+{
```
wooorm, line 1: Why move this? E.g., `bench` is also a package, but not placed in `packages/`.

When I now look in `packages/`, in part due to the location and also its name, `parse6-test-utils` sounds like something that’s published to npm and consumable for others.

Maybe a different name, or move it elsewhere?

wooorm (commented):
Here’s my review, coming in from nothing, of your changes.
**I did not yet look ts/js code, I’ll do that another time, I looked at everything else first**

bench/package.json
fb55: The `perf` benchmark compares the published module (from NPM) to the unreleased module. Because we are now using NPM workspaces, the original entry would have pulled in the unreleased module from the `packages` directory. By explicitly asking for the NPM module here, we work around that issue.

lerna.json
fb55: I haven't dealt with NPM workspaces in the context of a published module yet. My guess is no, lerna is no longer needed. I kept it around in case publishing becomes complicated for some reason.

package.json
fb55: Didn't know about this, thanks for flagging!

packages/parse5-html-rewriting-stream/docs/index.md
fb55: I moved a lot of the docs to JSDoc, which will make them accessible from editors. I haven't figured out how to document events yet, but that's the only reason why the original docs are still around.

@43081j — Do you have a suggestion of how to tackle events in JSDoc?

tsconfig.json
fb55: Node 12 needs es2019 as the target, according to https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping#node-12

packages/parse5-test-utils/data/parser-feedback/entities01.test
fb55: This was generated by the parser feedback generation script. Nothing fixed here, it just seems like that script wasn't run for a while and had some catching up to do.

packages/parse5-test-utils/data/parser-feedback/tests6.test
fb55: They were removed by the script. Seems like a change in the upstream test data that this is derived from.

packages/parse5-test-utils/package.json
fb55: Good point, I will move it back.

packages/parse5-test-utils/data/parser-feedback/entities01.test
fb55: As this isn't linked to any code changes, I'm open to reverting this. There were some actually breaking changes generated by the script, which I simply didn't include for now.

43081j:
ill try go through it this weekend.

for ongoing changes in the fork, is your plan to re-merge those back into here once they land? e.g. we still didn't manage to figure the strong typing of adapters out. so we should merge what we have but update them in a separate PR here

.eslintrc.json
43081j: i would recommend uvu but keep in mind it has no reporting functionality for if we want better test results in PRs (i.e. a format github integrations can understand). either way we should probably tackle that in a separate pr since this is already a monolith

fb55:
> for ongoing changes in the fork, is your plan to re-merge those back into here once they land? e.g. we still didn't manage to figure the strong typing of adapters out. so we should merge what we have but update them in a separate PR here

Let's do that in the main repo? I don't think there is much to gain by maintaining the fork.

.eslintrc.json
```diff
@@ -30,10 +30,26 @@
         "unicorn/no-array-reduce": "off",
         "unicorn/no-for-loop": "off",
         "unicorn/consistent-destructuring": "off",
-        "unicorn/prefer-switch": ["error", { "emptyDefaultCase": "do-nothing-comment" }],
-        "unicorn/number-literal-case": "off"
+        "unicorn/prefer-switch": ["error", { "emptyDefaultCase": "do-nothing-comment" }]
```
43081j, line 33: arguably we could just drop this dependency at some point since we pull it in for one minor rule we probably don't really need

.eslintrc.json
fb55: Agreed that Jest isn't great with ESM. I generally like the batteries-included approach, especially when we want to tackle issues such as #211.

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npm run pre-commit
```
43081j, line 4: probably worth upgrading husky at some point in a separate PR

fb55: Husky is already upgraded to the latest version as part of this PR. That's why this file is needed.

43081j: ah ignore me, this _is_ the new one!

packages/parse5-html-rewriting-stream/docs/index.md
43081j: events are a little odd in jsdoc, we can specify something like a `@fires` tag but that should also mean we define the event's type somewhere with `@event`. not entirely sure what convention vscode will look for though.

43081j:
yeah thats fair. ill open a new PR for the adapter types as a draft and see if i can figure it out. some middle ground exists somewhere im sure of it 😂 

if either of us can remember any unsolved issues i think we should open equivalents in this repo when we get time

packages/parse5-html-rewriting-stream/package.json
```diff
@@ -6,6 +6,9 @@
     "author": "Ivan Nikulin <ifaaan@gmail.com> (https://github.com/inikulin)",
     "contributors": "https://github.com/inikulin/parse5/graphs/contributors",
     "homepage": "https://github.com/inikulin/parse5",
+    "scripts": {
+        "build": "rm -r dist; tsc"
```
43081j, line 10: for sake of improving DX we may want to use rimraf or something so windows based contributors can get this building

.eslintrc.json
43081j: yeah i agree. if we did use uvu, uvu and c8/istanbul usually works alright and we would have coverage integration. but we wouldn't have any useful test results output (and unfortunately the uvu author seems to be against introducing that)

.eslintrc.json
```diff
@@ -2,7 +2,7 @@
     "env": {
         "es2020": true,
         "node": true,
-        "mocha": true
+        "jest": true
     },
     "extends": ["eslint:recommended", "prettier", "plugin:unicorn/recommended"],
```
43081j, line 7: at some point we should introduce `plugin:@typescript-eslint/recommended` but it'll need its own rather large cleanup PR i think since i suspect there's a fair amount of recommended rules we don't yet follow (but should)

fb55: That's done as an override below

.eslintrc.json
wooorm: Chromium has native coverage testing, which a small tool like [c8](https://github.com/bcoe/c8#readme) can expose.
Tape, brittle, uvu all support coverage

lerna.json
wooorm: All the npm commands now have a `--workspaces` flag, so you can use that to update versions (or do it manually), and use it to publish!

I’d say remove lerna?

packages/parse5-html-rewriting-stream/docs/index.md
wooorm: Not everyone uses TypeScript, and not everyone learns how projects work by reading the code.
I personally have an **really** hard time dealing with projects that don’t have docs. While the current docs are scattered, I’m :-1: on removing most/all docs.

packages/parse5-test-utils/data/parser-feedback/entities01.test
wooorm: Yeah, :+1: on that

packages/parse5-test-utils/data/parser-feedback/tests6.test
wooorm: weird

packages/parse5-test-utils/package.json
fb55: Done

packages/parse5-html-rewriting-stream/package.json
fb55: Thanks for flagging! I realised I could use `tic --build --clean` instead.

.eslintrc.json
43081j: i expected a lot more fixing up to get those rules passing, nice work

packages/parse5-htmlparser2-tree-adapter/lib/index.ts
```diff
@@ -0,0 +1,264 @@
...
+}
+
+export function setDocumentType(
+    document: Document,
+    name: string | null,
+    publicId: string | null,
+    systemId: string | null
+): void {
+    const data = doctype.serializeContent(name, publicId, systemId);
+    let doctypeNode = document.children.find(
```
43081j, line 114: you can drop the cast if you use type predicate: `find((node): node is ProcessingInstruction => ...)`

packages/parse5-htmlparser2-tree-adapter/lib/index.ts
```diff
@@ -0,0 +1,264 @@
...
+    }
+}
+
+export function adoptAttributes(recipient: Element, attrs: Attribute[]): void {
+    for (let i = 0; i < attrs.length; i++) {
+        const attrName = attrs[i].name;
+
+        if (typeof recipient.attribs[attrName] === 'undefined') {
+            recipient.attribs[attrName] = attrs[i].value;
+            recipient['x-attribsNamespace']![attrName] = attrs[i].namespace!;
```
43081j, line 185: typescript-eslint should warn on non-null assertions, it would be good if one day we can get rid of these. IIRC there's a lot of them, since the original source made a lot of assumptions about what has executed beforehand. just a note anyway, may open an issue so we don't forget

packages/parse5-htmlparser2-tree-adapter/lib/index.ts
```diff
@@ -0,0 +1,264 @@
...
+    return node.sourceCodeLocation as ElementLocation | null | undefined;
+}
+
+export function updateNodeSourceCodeLocation(node: Node, endLocation: Partial<ElementLocation>): void {
+    if (endLocation.endOffset != null) node.endIndex = endLocation.endOffset;
+
+    node.sourceCodeLocation = {
+        ...node.sourceCodeLocation,
+        ...endLocation,
+    } as any;
```
43081j, line 263: what happens when you don't cast this? would be good to get it working with a strong type

another future nice to have too: no-explicit-any rule and drop all the usages 🙏

packages/parse5-parser-stream/lib/index.ts
```diff
@@ -0,0 +1,105 @@
...
+    private lastChunkWritten = false;
+    private writeCallback: null | (() => void) = null;
+    private pausedByScript = false;
+
+    public parser: Parser<T>;
+    private pendingHtmlInsertions: string[] = [];
+    /** The resulting document node. */
+    public document: T['document'];
+
+    constructor(options?: ParserOptions<T>) {
```
43081j, line 39: another one for future: require accessibility modifiers on all methods/properties (will open an issue for these)

packages/parse5-parser-stream/lib/index.ts
```diff
@@ -0,0 +1,105 @@
...
+
+        this.writeCallback = callback;
+        this.parser.tokenizer.write(chunk, this.lastChunkWritten);
+        this._runParsingLoop();
+    }
+
+    // TODO [engine:node@>=16]: Due to issues with Node < 16, we are overriding `end` instead of `_final`.
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    override end(chunk?: any, encoding?: any, callback?: any): any {
```
43081j, line 62: it would be good to drop the `any`, its a shame this one of node's typical overloaded methods, but it could be this i think:

```ts
override end(cb?: () => void): Writable;
override end(chunk?: any, cb?: () => void): Writable;
override end(
  chunk?: any, // not sure we can get rid, the DT types need fixing so this is `unknown`
  encoding: BufferEncoding,
  cb?: () => void
): Writable;
```

the base one has those 3 overloads. i suppose these all being any is why the compiler allowed it

packages/parse5-parser-stream/lib/index.ts
43081j: opened #376

packages/parse5-parser-stream/lib/index.ts
```diff
@@ -0,0 +1,105 @@
...
+        this.lastChunkWritten = true;
+        super.end(chunk || '', encoding, callback);
+    }
+
+    //Scriptable parser implementation
+    private _runParsingLoop(): void {
+        this.parser.runParsingLoopForCurrentChunk(this.writeCallback, this._scriptHandler);
+    }
+
+    private _resume = (): void => {
```
43081j, line 72: don't suppose you know why these are all properties rather than methods? we should be able to pass a method as a script handler since it still matches the signature

packages/parse5-plain-text-conversion-stream/lib/index.ts
```diff
@@ -0,0 +1,39 @@
+import type { ParserOptions } from 'parse5/dist/parser/index.js';
```
43081j, line 1: whats going on with these paths? do we  not export them at the root? i.e. `'parse5'`

similarly, the tree adapter interface has no extension but these do

packages/parse5-parser-stream/lib/index.ts
fb55: If we passed methods below, we wouldn't get a binding of the `this` value. Using arrow functions works around that.

packages/parse5-parser-stream/lib/index.ts
fb55: Agreed that the code is currently super hacky. In Node < 16, trailing data would not be parsed properly, which is made me revert my previous beautiful `_final` override. The code is now equivalent to what was there before.

The [expiring comment](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/expiring-todo-comments.md) above will start failing once failing Node versions are no longer supported, which is why I'd leave things the way they are for now.

packages/parse5-htmlparser2-tree-adapter/lib/index.ts
fb55: I still have to add `startTag` and `endTag` in my types for `domhandler`.

lerna.json
fb55: Removed lerna

packages/parse5-test-utils/data/parser-feedback/webkit02.test
fb55: Reverted

packages/parse5-test-utils/data/parser-feedback/entities01.test
fb55: Done

packages/parse5-html-rewriting-stream/docs/index.md
fb55: The previous docs were already generated from TypeScript types, using `typedoc-plugin-markdown`. I would set up a website that makes them consumable outside of TypeScript, similar to http://cheerio.js.org

.eslintrc.json
fb55: Let's follow James' suggestion to tackle this in a separate PR. c8 seems super interesting, curious to see how that would look in a proper project.

packages/parse5/docs/index.md
fb55: The READMEs should definitely be extended a bit. Generated docs like this could also just live in their own space, eg. a website.

packages/parse5-htmlparser2-tree-adapter/lib/index.ts
43081j: ah i get you. maybe you can at least cast it as the thing it wants (which may require a prior `as unknown`, i.e. `as unknown as ActualType`).

and possibly leave a TODO on it? so we don't forget

43081j:
> 48 / 162 files viewed

getting there gradually 👀 

i'll open issues for stuff we want to do separately too (have already done a couple of linting etc). looks good so far though

packages/parse5-sax-parser/lib/index.ts
```diff
@@ -0,0 +1,312 @@
...
+            }
+        } while (!this.stopped && token.type !== TokenType.EOF);
+    }
+
+    protected _handleToken(token: Token): boolean {
+        if (token.type === TokenType.EOF) {
+            return true;
+        }
+
+        const { eventName, reshapeToken } = TOKEN_EMISSION_HELPERS[token.type];
```
43081j, line 181: this pattern (`TOKEN_EMISSION_HELPERS`) will always result in a disconnect of our types, i think. i had a think and im not sure there's a way to strongly type the value of a record based on its key.

in future i think we should probably change this to a switch so we gain the type inference from it. something as simple as:

```ts
switch (token.type) {
  case TokenType.START_TAG:
    this._emitToken('startTag', reshapeStartTagToken(token));
    break;
  // ...
}
```

would be more readable IMO too

packages/parse5-sax-parser/lib/index.ts
```diff
@@ -0,0 +1,312 @@
...
+    [TokenType.CHARACTER]: TEXT_EMISSION_HELPER,
+    [TokenType.NULL_CHARACTER]: TEXT_EMISSION_HELPER,
+    [TokenType.WHITESPACE_CHARACTER]: TEXT_EMISSION_HELPER,
+    [TokenType.HIBERNATION]: {
+        eventName: 'hibernation',
+        reshapeToken: (): Record<string, never> => ({}),
+    },
+};
+
+export interface SAXParser {
```
43081j, line 292: just so i understand, is this to cause typescript to merge the declarations and provide extra type information on top of the base class' `on` signature? i.e. this'll merge the interface here and the class above, so you don't have to override `on` to define the extra signatures.

43081j: and if that's the case, do you really need the base one at the bottom? won't that be inherited still?

package.json
fb55: I'm mostly weary of _helpful_ PRs that bump the patch or minor version, even though these versions are already covered by the range. Instead, I'd rather keep everything up to date. Plus, there are less package versions to ask for in bug reports (not really an issue here, as all of them are dev deps).

As long as no fixed versions are used, this shouldn't make life harder for `npm`.

packages/parse5-sax-parser/lib/index.ts
fb55: Good callout, updated with a switch.

packages/parse5-sax-parser/lib/index.ts
fb55: Your understanding is spot on, TS will merge these declarations and provide types for the events (see https://github.com/inikulin/parse5/runs/4775155629?check_suite_focus=true for an example of an error this caused when I had two token types swapped).

The order doesn't matter, happy to change it if there is a strong preference either way.

packages/parse5-htmlparser2-tree-adapter/lib/index.ts
fb55: Added TODO comments.

packages/parse5-sax-parser/lib/index.ts
43081j: what i mean is if you left the base one out completely, it'd still pick it up from the base class, wouldn't it? maybe we can drop it and keep the rest?

fb55: Not sure I follow. Could you elaborate?

fb55: There will be an error if the last one is not provided, even if it is defined in the base type 😕

fb55:
I just tackled the last issue that I encountered, which was that (1) builds still included assets they shouldn't have, and (2) none of the scripts worked. I fixed that by introducing TS project references, which also allows us to build the entire codebase with a single command.

wooorm (commented):
Thanks for your continued work. Glad that npm workspaces are working. I looked through everything except for TS code.

- The PR title and changelog presumably are not up to date?
- Docs are broken (they point to master here, which still includes everything). I’m still :-1: on removing all docs as it prevents us from cutting a release. I do find it acceptable if they will be added back in a different PR soon. As a separate point, which I believe you agree with (https://github.com/inikulin/parse5/pull/362#discussion_r782103554), I want to stress that I don‘t think autogenerated docs are enough.
- I’m still in favor of loosening semver ranges (https://github.com/inikulin/parse5/pull/362#discussion_r780649982). The argument you used for updating versions all the time was “*helpful* PRs”, to which I can say that, I use this style in 100s of projects for and have received very few of such useless PRs. I’m not going to block on this though, but it’s a strong preference.
- The major change, for dependents, seems to be mixins. I don‘t think the new system, or how to migrate, is documented?

It might be too much work, but when possible, I’d recommend pulling a few small things out of this PR when possible (e.g., “Switched the order of the formatted elements list”)

Future work:
- Switch from jest to something else, use C8
- Improve docs
- remove `parse5-fork` in the `parse5` org
- point to this repo in the `parse5` org

packages/parse5-sax-parser/lib/index.ts
43081j: well that sucks! 😂  so you can't specify overloads without also re-specifying the base signature? sad times, typescript

fb55:
> Thanks for your continued work. Glad that npm workspaces are working. I looked through everything except for TS code.

Thanks for doing so, and sorry for the delay to respond!

> - The PR title and changelog presumably are not up to date?

Added two points, but otherwise everything should be covered.

> - Docs are broken (they point to master here, which still includes everything). I’m still :-1: on removing all docs as it prevents us from cutting a release. I do find it acceptable if they will be added back in a different PR soon. As a separate point, which I believe you agree with (https://github.com/inikulin/parse5/pull/362#discussion_r782103554), I want to stress that I don‘t think autogenerated docs are enough.

The previous docs were generated using [`typedoc-plugin-markdown`](https://github.com/tgreyuk/typedoc-plugin-markdown), based on the `.tsd` file that used to be a part of this repo. They are moved back to TSDoc in this PR.

I Agree that docs are something that should be worked on. Let's discuss this in a new issue once this PR is merged.

> - I’m still in favor of loosening semver ranges (https://github.com/inikulin/parse5/pull/362#discussion_r780649982). The argument you used for updating versions all the time was “*helpful* PRs”, to which I can say that, I use this style in 100s of projects for and have received very few of such useless PRs. I’m not going to block on this though, but it’s a strong preference.

Sorry for not being clearer here. My main reason is that all versions in range have to be supported. By keeping versions at the latest, any future breakages are the upstream's responsibility.

I still have the [confusion](https://github.com/fb55/css-select/issues?q=is%3Aissue+is%3Aclosed+5.0.1) around security updates for `css-select` in mind as well, which is the point I made earlier.

> - The major change, for dependents, seems to be mixins. I don‘t think the new system, or how to migrate, is documented?

Removed mixins shouldn't require any changes for dependents. Classes no longer override their methods at runtime based on options. If anything, these changes should make monkey-patching more reliable.

> It might be too much work, but when possible, I’d recommend pulling a few small things out of this PR when possible (e.g., “Switched the order of the formatted elements list”)

Finding good delineation points is difficult, and I would like to go ahead as things are right now. For users reading the changelog, I do think that the changes here make sense: changes don't change any of the externally visible behaviour, only `@types/parse5` will have to be removed. I have started opening separate PRs for user-facing features and fixes.

Regarding _ Switched the order of the formatted elements list_: I swapped `push`/`pop` for `shift`/`unshift` in a few places where mostly the last 1 or 2 elements were used. Now, these will be at indices `0` and `1`, which leads to arguably more readable code. This should have been part of a separate PR, but as changes are split over several commits, reverting now is difficult 😒

> Future work:
> - Switch from jest to something else, use C8
> - Improve docs
> - remove `parse5-fork` in the `parse5` org
> - point to this repo in the `parse5` org

All of this makes a lot of sense to me.

fb55:
@43081j Have you had a chance to look further into this?

wooorm:
👋

> My main reason is that all versions in range have to be supported.

That’s how semver (ranges) are supposed to work though.
The trust-based system doesn’t always work (a dep shipping a breaking change in a patch/minor), but no unlocked semver range prevents it.
It is infrequent that I get an issue where someone doesn’t understand `npm update` (estimated one or two in 2 recent years across tons of repos).

> I still have the confusion around security updates for css-select in mind as well, which is the point I made earlier.

Reading the 5 issues you linked, the problem with `css-select` and `css-what` *seems* to be either a *pinned* dependency or everyone using v4 of `css-select` instead of the updated v5? I don’t see how the method I propose affects this.

I don’t think it’s worth optimizing for, when a sec vuln happens, 5 issues of people that don’t understand npm. Those folks will find issue trackers anyway for whatever reasons.
I’d rather optimize for a beautiful and readable commit log (I find it hard to read the [`css-select` commits](https://github.com/fb55/css-select/commits/master) because it’s all dependabot)

> Removed mixins shouldn't require any changes for dependents.

:+1:

> changes don't change any of the externally visible behaviour, only `@types/parse5` will have to be removed.

It’s hard for me to gauge whether that was the case. So, glad to hear.

43081j:
@fb55 really sorry things got busy so i haven't had much chance the last few days to be on github.

i'll try get through the rest of it by sunday

packages/parse5/lib/parser/formatting-element-list.ts
```diff
@@ -0,0 +1,154 @@
...
+            (entry) => entry.type === EntryType.Marker || this.treeAdapter.getTagName(entry.element) === tagName
+        );
+
+        return entry && entry.type === EntryType.Element ? entry : null;
+    }
+
+    getElementEntry(element: T['element']): ElementEntry<T> | null {
+        return this.entries.find(
+            (entry) => entry.type === EntryType.Element && entry.element === element
+        ) as ElementEntry<T> | null;
```
43081j, line 152: we should use a generic instead, `find<ElementEntry<T>>`

43081j (approved):

packages/parse5/lib/parser/formatting-element-list.ts
fb55: Good catch!

fb55:
Really happy to finally wrap this up. Thanks you so much for the reviews @wooorm and @43081j!

PR #378, opened 2022-01-13
Add serializeOuter function. Resolves #230.

### LINKED ISSUE #230: Add a serialization mode that mimics outerHTML

joeldenning:
See #230. This is an alternative to #347.

### DIFF
--- a/packages/parse5/lib/serializer/index.js
+++ b/packages/parse5/lib/serializer/index.js
@@ -36,6 +36,12 @@ export class Serializer {
         return this.html;
     }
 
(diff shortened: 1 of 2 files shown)


fb55:
I have done a bigger refactor of the serialisation code in #383. This added a `serializeOuter` method, and includes the test case from this PR (while naming you as a co-author). Thanks a lot @joeldenning for getting this started!

PR #383, opened 2022-01-18
Refactor & improve serializer

fb55:
Simplifies the serializer by removing the wrapping class, fixes mixed content (#333), adds the scripting flag (#332), and adds a `serializeOuter` method (#230, #378). Also fixes passing a `<template>` element to `serialize` resulting in an empty string.

The `parse5-serializer-stream` module should be deprecated after these changes.

### DIFF
--- a/packages/parse5-serializer-stream/lib/index.ts
+++ b/packages/parse5-serializer-stream/lib/index.ts
@@ -34,20 +34,11 @@ export class SerializerStream<T extends TreeAdapterTypeMap> extends Readable {
         super({ encoding: 'utf8' });
 
         this.serializer = new Serializer(node, options);
-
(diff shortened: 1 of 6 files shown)


kongxiangyan:
`serializeOuter` is really a convenient feature for just serializing a piece of  html works. I wonder know when it will be released in a official version. Or am I missed some approaches that can import the latest commit to my project❓

fb55:
I am trying to make a release happen – should only be another week or two.

wooorm:
@fb55 Maybe good to create a (pinned) issue tracking it and what needs to be done?

fb55:
@wooorm I've done a push to tackle the remaining items I want to see, and have opened PRs for them. Once these are concluded, I am ready to get a release out.

Issue #377, opened 2022-01-11
build improvement: disallow non-null assertions via ESLint


43081j:
We should enable the rule to disallow non-null assertions:

```ts
foo!.bar!.baz;
```

from what i remember, we have quite a lot of these scattered around because of the whole "bootstrap" concept. many methods make assumptions around which methods ran before them, meaning we "know" the value shouldn't be null, but _technically_ it could.

in all those cases, we should either add null checks (usually throwing if its null) or rework the bootstrapping to happen in the constructor (feel like @fb55 mentioned something similar to that a while ago).

PR #384, opened 2022-01-18
refactor(parser): Remove `_bootstrap` method

fb55:
The `_bootstrap` method leaves the parser in a broken-by-design state, which requires us to use non-null assertions to make TS happy. This PR removes this method, moving its arguments to the constructor.

This helps considerably with #377.

### DIFF
--- a/packages/parse5-parser-stream/lib/index.ts
+++ b/packages/parse5-parser-stream/lib/index.ts
@@ -1,5 +1,5 @@
 import { Writable } from 'node:stream';
-import { Parser, ParserOptions, defaultParserOptions } from 'parse5/dist/parser/index.js';
+import { Parser, ParserOptions } from 'parse5/dist/parser/index.js';
 import type { TreeAdapterTypeMap } from 'parse5/dist/tree-adapters/interface.js';
 import type { DefaultTreeAdapterMap } from 'parse5/dist/tree-adapters/default.js';
 
@@ -42,8 +42,7 @@ export class ParserStream<T extends TreeAdapterTypeMap = DefaultTreeAdapterMap> 
     constructor(options?: ParserOptions<T>) {
         super({ decodeStrings: false });
 
-        const opts = { ...defaultParserOptions, ...options };
-        this.parser = new Parser(opts);
+        this.parser = new Parser(options);
         this.document = this.parser.document;
     }
 
--- a/packages/parse5-parser-stream/lib/index.ts
+++ b/packages/parse5-parser-stream/lib/index.ts
@@ -1,5 +1,5 @@
 import { Writable } from 'node:stream';
-import { Parser, ParserOptions } from 'parse5/dist/parser/index.js';
+import { Parser, ParserOptions, defaultParserOptions } from 'parse5/dist/parser/index.js';
 import type { TreeAdapterTypeMap } from 'parse5/dist/tree-adapters/interface.js';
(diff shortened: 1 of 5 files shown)


wooorm:
I’m a heavy user of the internals of parse5, [including `_bootstrap`](https://github.com/syntax-tree/hast-util-raw/blob/83e93749718eb5846d6bc70aaac545a12c887e59/lib/index.js#L200-L205). I hope there will be alternatives?

fb55:
The arguments of `_bootstrap` are now passed to the constructor. Please let me know if that is acceptable!

The rationale behind this change: Afaict the original intention of `_bootstrap` was to reuse the same parser instance multiple times. That goal isn't quite met, as all of the child classes are re-instantiated anyhow. By removing `_bootstrap`, we eliminate a state of the parser that is broken by design (instantiated but not bootstrapped).

wooorm:
> The arguments of `_bootstrap` are now passed to the constructor. Please let me know if that works for you!

That seems fine. I wanted to stress that I’m using the internals of parse5 to do some magic. While hacky and not pretty, I’d rather not see all internals hidden away.
This change seems unrelated, so that’s fine!

packages/parse5/lib/parser/index.ts
```diff
@@ -82,7 +82,7 @@ export interface ParserOptions<T extends TreeAdapterTypeMap> {
      *
      *  @default `true`
      */
-    scriptingEnabled?: boolean | undefined;
+    scriptingEnabled?: boolean;
```
wooorm, line 85: Is it intentional that the value can’t be `undefined`, but that the field can be missing?

packages/parse5-parser-stream/lib/index.ts
```diff
@@ -42,10 +42,9 @@ export class ParserStream<T extends TreeAdapterTypeMap = DefaultTreeAdapterMap>
     constructor(options?: ParserOptions<T>) {
         super({ decodeStrings: false });
 
-        this.parser = new Parser(options);
-
-        this.document = this.parser.treeAdapter.createDocument();
-        this.parser._bootstrap(this.document, null);
+        const opts = { ...defaultParserOptions, ...options };
+        this.parser = new Parser(opts);
```
wooorm, line 46: Why aren’t defaults handled in `Parser`?

fb55: Done!

packages/parse5/lib/parser/index.ts
fb55: These values can still be explicitly undefined — I am honestly not sure why the `| undefined` was here in the first place (this was copied from the previous @types package). My suspicion would be an older TS version?

The only difference is that now `Required<>` works with the options.

wooorm: I know that DefinitelyTyped was adding explicit `| undefined` to types, when they had `?`. They‘re planning to pull them apart: https://github.com/DefinitelyTyped/DefinitelyTyped/commit/b24e5f670af4b29aa131dedc1db578d08d40c028

fb55: Ahh, thanks for digging this up. Looks like the option is now called `exactOptionalPropertyTypes` and is explicitly opt-in (no longer enabled with `strict`): https://github.com/microsoft/TypeScript/pull/44626

My intuition is that users enabling this flag wouldn't want to be able to pass `undefined` here.

Issue #285, opened 2018-11-18
Exposing parentElement on treeAdapter.createElement 


pmdartus:
While implementing support for [custom elements in JSDOM](https://github.com/jsdom/jsdom/issues/1030#issuecomment-437083550), I am facing the following limitation. The [create an element for a token](https://html.spec.whatwg.org/multipage/parsing.html#create-an-element-for-the-token) algorithm use the *intended parent* to retrieve the associated document for a newly created node. 

In it's current implementation the JSDOM treeAdapter shares the [same document](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/browser/parse5-adapter-parsing.js#L11-L13) for all the created node. However this approach doesn't work in the context of template element, since the HTMLTemplateElment content is associated with an [inert template document](https://html.spec.whatwg.org/multipage/scripting.html#associated-inert-template-document).

What do you think about extended the current createElement signature to pass the intededParent as parameter?

```ts
// The intendedParent can be undefined when this hook is invoked for the root html element.
createElement(tagName: string, namespaceURI: string, attrs: Attribute[], intendedParent?: Element): Element;
```

An alternative viable approach would be to expose in some way the [current element stack](https://github.com/inikulin/parse5/issues/237).

inikulin:
Sorry for this taking that long. I'm alright with adding this change and will be happy to merge a PR for it if you mind to create one.

fb55:
The alternative viable approach has been implemented in #385.

PR #385, opened 2022-01-18
feat(parser): Add hooks for stack events

### LINKED ISSUE #237: Need more insight into the stack of open elements
I've tried to come up with some alternative but it seems that to implement the [style element](https://html.spec.whatwg.org/multipage/semantics.html#the-style-element) properly, we need some notification: it requires action when

> - The element is popped off the stack of open elements of an HTML parser or XML parser.
>
> - The element is not on the stack of open elements of an HTML parser or XML parser, and it becomes connected or disconnected.

Do you have any ideas on how we could implement this in a parse5 adapter? It's currently causing a pretty bad regression in jsdom ever since we switched from the htmlparser2 adapter, with bad consequences for our consumers.

fb55:
JSDOM currently monkey-patches parse5 to get these events. With this change, that is no longer necessary.

Fixes #237, #285.

### DIFF
--- a/packages/parse5/lib/parser/index.test.ts
+++ b/packages/parse5/lib/parser/index.test.ts
@@ -117,11 +117,15 @@ describe('parser', () => {
             assert.ok(isElementNode(htmlElement));
             const bodyElement = htmlElement.childNodes[1];
             assert.ok(isElementNode(bodyElement));
+            // Expect 5 opened elements; in order: html, head, body, and 2x p
             expect(onItemPush).toHaveBeenCalledTimes(5);
             expect(onItemPush).toHaveBeenNthCalledWith(1, htmlElement);
             expect(onItemPush).toHaveBeenNthCalledWith(3, bodyElement);
+            // The last opened element is the second p
             expect(onItemPush).toHaveBeenLastCalledWith(bodyElement.childNodes[1]);
+            // The second p isn't closed, plus we never pop body and html. Alas, only 2 pop events (head and p).
             expect(onItemPop).toHaveBeenCalledTimes(2);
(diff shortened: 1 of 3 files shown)


packages/parse5/lib/parser/index.test.ts
```diff
@@ -98,4 +100,29 @@ describe('parser', () => {
...
+            const htmlElement = document.childNodes[0];
+            assert.ok(isElementNode(htmlElement));
+            const bodyElement = htmlElement.childNodes[1];
+            assert.ok(isElementNode(bodyElement));
+            expect(onItemPush).toHaveBeenCalledTimes(5);
+            expect(onItemPush).toHaveBeenNthCalledWith(1, htmlElement);
+            expect(onItemPush).toHaveBeenNthCalledWith(3, bodyElement);
+            expect(onItemPush).toHaveBeenLastCalledWith(bodyElement.childNodes[1]);
+            expect(onItemPop).toHaveBeenCalledTimes(2);
+            expect(onItemPop).toHaveBeenLastCalledWith(bodyElement.childNodes[0], bodyElement);
```
wooorm, line 125: Looks a bit like magic numbers. Perhaps good to comment why 5, 1, 3, 2?

wooorm (commented):
Can you reference where this was requested? The reason this is added?

fb55:
> Can you reference where this was requested? The reason this is added?

Ahh, good catch. This was requested in #237, plus again in #285.

packages/parse5/lib/parser/index.test.ts
fb55: Added some comments.

PR #397, opened 2022-02-07
refactor(tokenizer): Re-use object for current location

fb55:
Previously, `ctLoc` was overwritten on every new character. This change should lead to a great reduction in allocations when using location information.

### DIFF
--- a/packages/parse5/lib/tokenizer/index.ts
+++ b/packages/parse5/lib/tokenizer/index.ts
@@ -231,6 +231,7 @@ export class Tokenizer {
         this.addLocationInfo = !!options.sourceCodeLocationInfo;
         this.onParseError = options.onParseError ?? null;
         this.preprocessor = new Preprocessor(options);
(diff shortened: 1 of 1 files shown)


packages/parse5/lib/tokenizer/index.ts
```diff
@@ -925,7 +953,7 @@ export class Tokenizer {
     //------------------------------------------------------------------
     private _stateData(cp: number): void {
         this.preprocessor.dropParsedChunk();
-        this.ctLoc = this._getCurrentLocation();
+        this.updateStartLocation();
```
wooorm, line 956: feels a bit weird having to do this book-keeping — its existence seems to indicate that the data might be out of date and before you access it you’d have to call this function?

fb55: I played around with this some more. Turns out we can actually get rid of location updates in the *DATA states altogether: #402

fb55:
Closing in favour of #402

PR #402, opened 2022-02-11
refactor(tokenizer): Use explicit offsets for locations

fb55:
Supersedes #397

Previously, `ctLoc` was overwritten on every new character. #397 attempted to improve this by maintaining the locations in an updated object, and creating copies when needed.

With this PR, we instead use explicit offsets when creating tokens. There is no more need to update locations; instead, we exploit the fact that a *DATA state is entered only after emitting the previous token.

This change should lead to a great reduction in allocations when using location information.

### DIFF
--- a/packages/parse5/lib/tokenizer/index.ts
+++ b/packages/parse5/lib/tokenizer/index.ts
@@ -242,6 +242,7 @@ export class Tokenizer {
 
     private currentAttrLocation: Location | null = null;
 
+    // NOTE: `offset` may never run across line boundaries.
     private getCurrentLocation(offset: number): Location | null {
         if (!this.addLocationInfo) {
             return null;
--- a/packages/parse5/lib/tokenizer/index.ts
+++ b/packages/parse5/lib/tokenizer/index.ts
@@ -225,9 +225,6 @@ export class Tokenizer {
     private currentToken: Token | null = null;
     private currentAttr: Attribute = { name: '', value: '' };
 
-    // NOTE: Doctypes tokens are created with several different offset. We keep track of the moment a doctype *might* start here.
-    private doctypeStartLoc: Location | null = null;
-
     private addLocationInfo;
     private onParseError;
 
@@ -370,7 +367,7 @@ export class Tokenizer {
             forceQuirks: false,
             publicId: null,
             systemId: null,
-            location: this.doctypeStartLoc,
+            location: this.ctLoc,
         };
     }
 
@@ -1961,7 +1958,8 @@ export class Tokenizer {
             this._createCommentToken($$.DASH_DASH.length + 1);
             this.state = State.COMMENT_START;
         } else if (this._consumeSequenceIfMatch($$.DOCTYPE, false)) {
-            this.doctypeStartLoc = this.getCurrentLocation($$.DOCTYPE.length + 1);
+            // NOTE: Doctypes tokens are created without fixed offsets. We keep track of the moment a doctype *might* start here.
+            this.ctLoc = this.getCurrentLocation($$.DOCTYPE.length + 1);
             this.state = State.DOCTYPE;
         } else if (this._consumeSequenceIfMatch($$.CDATA_START, true)) {
             if (this.allowCDATA) {
--- a/packages/parse5/lib/tokenizer/index.ts
+++ b/packages/parse5/lib/tokenizer/index.ts
@@ -220,7 +220,7 @@ export class Tokenizer {
 
     private consumedAfterSnapshot = -1;
 
-    private ctLoc: Location | null;
+    private currentLocation: Location | null;
     private currentCharacterToken: CharacterToken | null = null;
     private currentToken: Token | null = null;
     private currentAttr: Attribute = { name: '', value: '' };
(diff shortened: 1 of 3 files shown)


packages/parse5/lib/tokenizer/index.ts
```diff
@@ -220,17 +220,22 @@ export class Tokenizer {
 
     private consumedAfterSnapshot = -1;
 
+    private ctLoc: Location | null;
     private currentCharacterToken: CharacterToken | null = null;
     private currentToken: Token | null = null;
     private currentAttr: Attribute = { name: '', value: '' };
 
+    // NOTE: Doctypes tokens are created with several different offset. We keep track of the moment a doctype *might* start here.
```
wooorm, line 228: Weird? Should examples of differences perhaps be documented here?

packages/parse5/lib/tokenizer/index.ts
```diff
@@ -220,17 +220,22 @@ export class Tokenizer {
...
+    private doctypeStartLoc: Location | null = null;
+
     private addLocationInfo;
     private onParseError;
 
     constructor(options: { sourceCodeLocationInfo?: boolean; onParseError?: ParserErrorHandler | null }) {
         this.addLocationInfo = !!options.sourceCodeLocationInfo;
         this.onParseError = options.onParseError ?? null;
         this.preprocessor = new Preprocessor(options);
+        this.ctLoc = this.getCurrentLocation(-1);
```
wooorm, line 238: not `0`?

packages/parse5/lib/tokenizer/preprocessor.ts
```diff
@@ -17,7 +17,8 @@ const DEFAULT_BUFFER_WATERLINE = 1 << 16;
 export class Preprocessor {
     public html = '';
     private pos = -1;
-    private lastGapPos = -1;
+    // NOTE: Initial `lastGapPos` is -2, to ensure `col` on initialisation is 0
+    private lastGapPos = -2;
```
wooorm, line 21: `-2` seems weird, `undefined` or so perhaps? Or would that make the types too complex?

packages/parse5/lib/tokenizer/index.ts
fb55: It just occurred to me that the `ctLoc` property can be used here; updated accordingly.

The issue with doctypes is that they don't start at fixed offsets. Eg. whitespace can push the token creation further back.

wooorm: how is that different from elements, or anything else, that is also pushed back by whitespace?

fb55: Doctype tokens are a bit weird in the way they are created in the spec. All other tokens are created as soon as you can be certain that a specific type of token is needed. With Doctypes, this is pushed back until we get the beginning of something that needs to be saved. Eg. `<!doctype         html>` would only have its token created on the `h` of `html`.

packages/parse5/lib/tokenizer/preprocessor.ts
fb55: That would complicate things quite a bit. The issue here is the check `Number(this.lastGapPos !== this.pos)` below. `-2` is weird, but does make things work as expected.

packages/parse5/lib/tokenizer/index.ts
fb55: The doubly negated offset makes things a bit difficult here. An offset of `+1` means "start one character ago". `-1` means "start one character after our last seen one".

The preprocessor is initiated at position `-1`. We access the location here before we read anything, so have to give a negative offset of `-1` to get the position of the first character in the input. The same is done again in `_emitCurrentToken`.

I have played with flipping the sign of `getCurrentLocation`'s `offset`, but then other places are hard to read.

packages/parse5/lib/tokenizer/index.ts
wooorm: Do you have a link to the spec that mentions this behavior?
I don’t understand, from the tokenizing chapter, what makes `DOCTYPE`, `[CDATA[`, or `--` different?

packages/parse5/lib/tokenizer/preprocessor.ts
wooorm: `Infinity` (or `-Infinity`) perhaps?

packages/parse5/lib/tokenizer/index.ts
fb55: The [Before DOCTYPE name state](https://html.spec.whatwg.org/multipage/parsing.html#before-doctype-name-state) is probably the best example here. The token is only created if a non-whitespace character is found.

This is in contrast to the [Markup declaration open state](https://html.spec.whatwg.org/multipage/parsing.html#markup-declaration-open-state), which creates comment tokens after reading a specific number of characters. This allows us to set specific offsets. The same applies to the [Tag open state](https://html.spec.whatwg.org/multipage/parsing.html#tag-open-state) and all of the `* end tag open state`s. These are all the states in which non-doctype tokens are created.

packages/parse5/lib/tokenizer/preprocessor.ts
fb55: `-Infinity` seems even weirder to me tbh 😄 

`-2` still has a resemblance to the _not found_ return value of `indexOf`, with `-Infinity` I am taken aback by the sudden floating point number for something that is supposed to be an index.

PR #404, opened 2022-02-11
refactor(tokenizer): Introduce events

fb55:
Update the tokenizer to produce events. See #403 for details.

Issue #403, opened 2022-02-11
Proposal: Replace the token queue with an event-handler system


fb55:
### Why

The token queue adds a level of indirection that makes it harder to fix some issues. Eg. #292 is easy to fix once the token queue is gone. Also, debugging is currently complicated, as stack traces end at the token queue.

With the queue gone, stack traces will point at the corresponding line in the tokenizer. V8 will be able to optimise more aggressively; in my branch combining all of the changes, I see a ~15% performance increase using `htmlparser-benchmark`.

### Game plan

1. Update the tokenizer to produce events. There will be a `QueuedTokenizer` class that wraps around the tokenizer, which provides an interface for the parser. Opened as #404
2. Invert event processing in the parser. The parser currently first checks the insertion mode, and then the token type. By inverting this (checking first the token type, then the insertion mode), we prepare the parser to accept the events from (1). Opened as #405
3. Tie everything together. Have the updated parser from (2) consume the tokenizer events from (1). Opened as #419

(1) and (2) do not depend on one-another and can be merged independently.

cc @wooorm @43081j

PR #405, opened 2022-02-11
refactor(parser): Invert event processing

fb55:
See #403 for details.

The parser currently first checks the insertion mode, and then the token type. By inverting this (checking first the token type, then the insertion mode), we prepare the parser to accept the events from #404.

This change also allowed foreign content checks to be greatly simplified.

### DIFF
--- a/packages/parse5/lib/tokenizer/index.ts
+++ b/packages/parse5/lib/tokenizer/index.ts
@@ -213,7 +213,7 @@ export class Tokenizer {
      * Indicates that the current adjusted node exists, is not an element in the HTML namespace,
      * and that it is not an integration point for either MathML or HTML.
      *
-     * @see {@link https://html.spec.whatwg.org/#tree-construction}
+     * @see {@link https://html.spec.whatwg.org/multipage/parsing.html#tree-construction}
      */
     public inForeignNode = false;
     public lastStartTagName = '';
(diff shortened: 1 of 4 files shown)


fb55:
One thing I am currently unhappy about is the `tokenizer.inForeignNode` and `currentNotInHTML` naming inside of the parser.

- `currentNotInHTML` indicates that the current node is not an element in the HTML namespace.
- `tokenizer.inForeignNode` (fka. `tokenizer.allowCDATA`) indicates that the current node is not an element in the HTML namespace, and that it is not an integration point for either MathML or HTML. (See https://html.spec.whatwg.org/#tree-construction)

If someone can come up with better names, I'm am more than happy to take them.

packages/parse5/lib/tokenizer/index.ts
```diff
@@ -209,7 +209,13 @@ export class Tokenizer {
 
     private tokenQueue: Token[] = [];
 
-    public allowCDATA = false;
+    /**
+     * Indicates that the current adjusted node exists, is not an element in the HTML namespace,
+     * and that it is not an integration point for either MathML or HTML.
+     *
+     * @see {@link https://html.spec.whatwg.org/#tree-construction}
```
wooorm, line 216: link to multipage perhaps, otherwise it’s too slow?

wooorm (commented):
Isn’t this a deviation from how the spec is written? Seems like that’ll cause future problems.

fb55:
> Isn’t this a deviation from how the spec is written? Seems like that’ll cause future problems.

I don't think this will make maintenance noticeably harder. parse5 already has a split for specific token handlers that doesn't exist in the spec. The splitting point was moved up by a step, which shouldn't make it harder to map the spec to the code.

PR #418, opened 2022-02-26
feat: Add CommonJS export for parse5 module

### LINKED ISSUE #379: Add CommonJS output
As discussed in #351, we should make parse5 a hybrid module, with both ESM and CommonJS entry points.

fb55:
Fixes #379 

This mostly follows the pattern from [SenseDeep](https://www.sensedeep.com/blog/posts/2021/how-to-create-single-source-npm-module.html).

cc @43081j @wooorm


package.json
```diff
@@ -24,7 +24,9 @@
         "typescript": "^4.5.5"
     },
     "scripts": {
-        "build": "tsc --build packages/* test",
+        "build": "npm run build:es && npm run build:cjs",
+        "build:es": "tsc --build packages/* test",
+        "build:cjs": "tsc -p packages/parse5/tsconfig.cjs.json && echo '{\"type\":\"commonjs\"}' > packages/parse5/dist/cjs/package.json",
```
wooorm, line 29:
```suggestion
        "build": "npm run build:esm && npm run build:cjs",
        "build:esm": "tsc --build packages/* test",
        "build:cjs": "tsc -p packages/parse5/tsconfig.cjs.json && echo '{\"type\":\"commonjs\"}' > packages/parse5/dist/cjs/package.json",
```

`esm` is used a bunch everywhere. And both 3 characters means it aligns nicely :)

fb55: Good recommendation! 😄

wooorm (commented):
this looks good, can the CI be fixed though?

fb55:
The CI is fixed in #419

package.json
```diff
@@ -24,7 +24,9 @@
         "typescript": "^4.5.5"
     },
     "scripts": {
-        "build": "tsc --build packages/* test",
+        "build": "npm run build:esm && npm run build:cjs",
+        "build:esm": "tsc --build packages/* test",
+        "build:cjs": "tsc -p packages/parse5/tsconfig.cjs.json && echo '{\"type\":\"commonjs\"}' > packages/parse5/dist/cjs/package.json",
```
43081j, line 29: does this make node think `dist/cjs/` is its own package?

packages/parse5/package.json
```diff
@@ -27,7 +27,15 @@
         "serialize"
     ],
     "license": "MIT",
-    "main": "dist/index.js",
+    "main": "dist/cjs/index.js",
```
43081j, line 30: wouldn't we still want the default to be ESM? so the only way to reach any commonjs entrypoints is by reading the package exports

package.json
fb55: This will make `require('parse5')` work in Node versions with ESM support.

Because the files are plain `.js` files, Node will look at the closest `package.json` file to determine if the code is ESM or not. We create a `package.json` for the CJS code, to make sure requiring it works as expected.

packages/parse5/package.json
fb55: Node versions with ESM support will not read the `main` property if `exports` are specified. That means this field is only for non-ESM Node versions, which should use the CJS version.

package.json
fb55: I can't think of any side-effects, but definitely something worth keeping in mind for future bug-reports. I adopted the pattern from [this article](https://www.sensedeep.com/blog/posts/2021/how-to-create-single-source-npm-module.html) from SenseDeep, which is used in several published (but not very popular) modules.

fb55:
Thanks for the review y'all!

packages/parse5/package.json
```diff
@@ -27,7 +27,15 @@
...
+    "module": "dist/index.js",
+    "types": "dist/index.d.ts",
+    "exports": {
+        ".": {
+            "import": "dist/index.js",
+            "require": "dist/cjs/index.js"
+        }
+    },
```
milahu, line 38: this breaks [rehype/test/parse-error.js](https://github.com/rehypejs/rehype/blob/620bb9cdf53a4dbe70adfca8d66c8824eb3f513f/test/parse-error.js#L7)

```js
import p5errors from 'parse5/lib/common/error-codes.js'
```

wontfix in rehype, as [ERR](https://github.com/inikulin/parse5/blob/075be2e712644410e1ceae67ebabad958bc0b66e/packages/parse5/lib/common/error-codes.ts#L9) is not exported by parse5

&rarr; export as ErrorCodes?

https://nodejs.org/api/esm.html

> module files within packages can be accessed by appending a path to the package name **unless the package's [package.json](https://nodejs.org/api/packages.html#nodejs-packagejson-field-definitions) contains an ["exports"](https://nodejs.org/api/packages.html#exports) field, in which case files within packages can only be accessed via the paths defined in ["exports"](https://nodejs.org/api/packages.html#exports).**

43081j: the package exports are right IMO. if we have consumers of the error codes enum, we should probably just export it top-level along with `ParserError`, possibly as a useful name like you said (`ErrorCodes`).

i.e. in `index.ts`, `export {ERR as ErrorCodes}` or something

PR #419, opened 2022-02-27
refactor(parser): Consume tokenizer events

### LINKED ISSUE #403: Proposal: Replace the token queue with an event-handler system

fb55:
Fixes #403 

Final PR for #403. Now that everything is prepared, we can switch the parser to consuming tokenizer events.

This required an update for the parser feedback test generation script, which now has to consume events as well.


fb55:
`master` is currently broken due to a misalignment of the previous two commits. This PR will resolve the issue.

packages/parse5/lib/tokenizer/tokenizer-location-info.test.ts
```diff
@@ -11,22 +10,17 @@ interface LocationInfoTestCase {
 }
 
 /** Receives events and immediately compares them against the expected values. */
-class LocationInfoHandler extends SinglePathHandler {
```
fb55, line 14: `SinglePathHandler` was merged into `LocationInfoHandler`.

PR #429, opened 2022-03-02
refactor: Use handler for `OpenElementStack`

fb55:
Now that the tokenizer is using the delegate pattern, it only makes sense to update the `OpenElementStack` to follow the same pattern.


43081j (approved):
makes sense to me :+1:

PR #430, opened 2022-03-02
fix(tokenizer): No parse error on attribute quirk

fb55:
Turns out a bug I _"fixed"_ in my fork (in https://github.com/inikulin/parse5/commit/70c1935552e90f23e2cbf4d0e2ce60d1da0b7e49) was actually the specced behaviour. This PR reintroduces the old behaviour.

Thanks to @untitaker for flagging this in https://github.com/html5lib/html5lib-tests/pull/144.


wooorm (approved):

fb55:
Thanks for the review @wooorm!

PR #432, opened 2022-03-03
fix(tokenizer): Drop chunks after emitting tokens

### LINKED ISSUE #292: Rewriter: text content longer than 65536 chars is truncated

fb55:
Fixes #292, #357

We would sometimes drop chunks before the corresponding tokens were processed by the rewriting stream. This led to #292. Now, we will drop buffers after emitting events.

BREAKING CHANGE: The SAX parser and rewriting stream will now emit text events when the underlying buffer will be dropped right after the event.

@qnighy I've adopted your test from #357 — hope that's okay! I'd also love it if you could make sure this works for you.

### DIFF
--- a/packages/parse5-html-rewriting-stream/test/rewriting-stream.test.ts
+++ b/packages/parse5-html-rewriting-stream/test/rewriting-stream.test.ts
@@ -306,8 +306,8 @@ describe('RewritingStream', () => {
         assert.throws(() => stream.write(buf), TypeError);
     });
 
-    it('Regression - RewritingStream - should pass long text correctly (GH-292)', (done) => {
-        const source = 'a'.repeat(65_540);
+    it('Should pass long text correctly (GH-292)', (done) => {
+        const source = 'a'.repeat((1 << 16) + 1);
         const parser = new RewritingStream();
         let output = '';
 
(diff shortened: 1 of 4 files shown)


packages/parse5/lib/tokenizer/index.ts
```diff
@@ -536,6 +542,7 @@ export class Tokenizer {
             if (this.currentCharacterToken.type !== type) {
                 this.currentLocation = this.getCurrentLocation(0);
                 this._emitCurrentCharacterToken(this.currentLocation);
```
fb55, line 544: We can't drop the parsed chunk in `_emitCurrentCharacterToken`, as it is called from `prepareToken` and would remove the token right afterwards.

packages/parse5-html-rewriting-stream/test/rewriting-stream.test.ts
```diff
@@ -305,4 +305,48 @@ describe('RewritingStream', () => {
...
+    it.only('Should emit comment after text correctly', (done) => {
+        const source = `${'a'.repeat((1 << 16) - 3)}<!-- foo -->`;
+        const parser = new RewritingStream();
+        let output = '';
+
+        parser.on('data', (data) => {
+            output += data.toString();
+        });
+
+        parser.once('close', () => {
```
fb55, line 340: Switching this line to the `finish` event will make the test fail, due to the stream buffer waterline being hit. Other tests have been updated to also use the `close` event, for consistency.

wooorm (commented):
not my area of expertise in this code base. But if the test was broken before and works now, :+1:

qnighy:
Thanks a lot for the fix!

PR #434, opened 2022-03-04
fix(rewriting-stream): Don't escape text in special tags

### LINKED ISSUE #339: RewritingStream incorrectly escaping contents of <script> tags via emitText()
This is related to but distinct from #333

The `RewritingStream.emitText()` method always calls escapeString on the text:
https://github.com/inikulin/parse5/blob/9c683e154b8bbdc94237580680a934348406d1a2/packages/parse5-html-rewriting-stream/lib/index.js#L79-L81

This causes it to break JavaScript. For example:

```html
<script type="text/javascript">
(function() {
  if ('PerformanceObserver' in window && 'PerformancePaintTiming' in window) {
    // ...
```
ends up as 

```html
<script type="text/javascript">
(function() {
  if ('PerformanceObserver' in window &amp;&amp; 'PerformancePaintTiming' in window) {
    // ...
```
and then the browser's JS parser chokes with a syntax error when it hits the `&amp;&amp;`.

It should probably do something similar to what _serializeTextNode does:

https://github.com/inikulin/parse5/blob/9c683e154b8bbdc94237580680a934348406d1a2/packages/parse5/lib/serializer/index.js#L136-L149

(Or, whatever it ends up doing after the resolution of #333)

For a short-term workaround, I'm going to just call ~~`stream.push()`~~ `stream.emitRaw()` when inside of a <script> tag.

fb55:
Fixes #339

__Question:__ This allows users to _leak_ data from tags. Eg. inserting the text `</script> foo` would close the current script tag. There is no good solution that works for all special tags, and I'm not even sure this is an issue.

cc @nfriedly — I'd love to hear your take on this.

### DIFF
--- a/packages/parse5-html-rewriting-stream/lib/index.ts
+++ b/packages/parse5-html-rewriting-stream/lib/index.ts
@@ -1,6 +1,6 @@
 import type { Location } from 'parse5/dist/common/token.js';
 import { SAXParser, EndTag, StartTag, Doctype, Text, Comment, SaxToken } from 'parse5-sax-parser';
-import { escapeString } from 'parse5/dist/serializer/index.js';
+import { hasUnescapedText, escapeString } from 'parse5/dist/serializer/index.js';
 
 /**
  * Streaming [SAX](https://en.wikipedia.org/wiki/Simple_API_for_XML)-style HTML rewriter.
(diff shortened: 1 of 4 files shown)


nfriedly:
Looks good to me. Thank you!

fb55:
cc @alan-agius4 @clydin as the last contributors to Angular's rewriting stream code:

The Angular team uses the rewriting stream; as probably the main consumer of this: Is the possibility of text escaping its container (see question above) a concern for you?

alan-agius4:
@fb55, I don’t think this is a problem at all.

fb55:
> I don’t think this is a problem at all.

Awesome, thanks for confirming.

PR #435, opened 2022-03-05
chore: Add funding information

fb55:
Adds funding information to the repo & packages. I don't know if sponsorships are enabled; if not, we'll have to reach out to Ivan.

@wooorm I've added your GitHub Sponsors, let me know if this is wanted.

@43081j Do you have something you'd like to add here?

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,2 +1,2 @@
 open_collective: parse5
-github: [fb55, wooorm]
+github: [fb55]
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,2 @@
+open_collective: parse5
+github: [fb55, wooorm]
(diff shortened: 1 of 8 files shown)


wooorm:
Cool!

Is the `?sponsor=1` working for y’all? It didn’t for me just now. I know Sindre used to do it/does it?

I’m also fine not being named. If your plan is to actively take care of most of the work to maintain this, I think it should be focussed on you.

If this is more a thing where several folks casually do things when they have the time, no pressure, then it makes sense adding folks that are (at the time) active?

43081j:
maybe because it doesn't yet exist in master? unless i misunderstood how its meant to work :D

fb55:
> I’m also fine not being named. If your plan is to actively take care of most of the work to maintain this, I think it should be focussed on you.

Makes sense. In a similar spirit, feel free to remove me should I ever drop off the picture.

> maybe because it doesn't yet exist in master? unless i misunderstood how its meant to work :D

That's my understanding as well. There is a chance that Ivan will have to enable Sponsors before this will be visible, but we'll know after merging.

PR #440, opened 2022-03-07
refactor: Make location info parser tests work

fb55:
I realised during the original refactor that `generateLocationInfoParserTests` didn't call some of the functions is was supposed to call. Digging into this now, it turned out that it was calling `escapeString` on the input before processing anything. That meant we would receive a single long string with all of the document's content, and only assert a single time that the received string matched the input.

Now, we properly walk the tree (depth first, from the deepest node), and test nodes along the way.

#436 was found during this effort.

Other changes necessary to make this work:
- `serializeOuter` was updated to serialise any kind of node.
- Removed `null` from document type functions (value was never used)
- **BREAKING:** Before, we would always add an empty `SYSTEM` identifier to the data prop of document types of the htmlparser2 adapter. This was removed; ie. `<!DOCTYPE html "">` is now `<!DOCTYPE html>`.

### DIFF
--- a/test/utils/generate-location-info-parser-tests.ts
+++ b/test/utils/generate-location-info-parser-tests.ts
@@ -4,7 +4,6 @@ import { TreeAdapter, TreeAdapterTypeMap } from 'parse5/dist/tree-adapters/inter
 import * as assert from 'node:assert';
 import * as fs from 'node:fs';
 import * as path from 'node:path';
-import { escapeString } from 'parse5/dist/serializer/index.js';
 import * as parse5 from 'parse5/dist/index.js';
 import {
     removeNewLines,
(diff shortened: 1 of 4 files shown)


test/utils/generate-location-info-parser-tests.ts
```diff
@@ -113,42 +112,44 @@ export function generateLocationInfoParserTests(
...
+                              treeAdapter.getDocumentTypeNodeSystemId(node)
+                          )}>`
+                        : parse5.serializeOuter(node, { treeAdapter });
 
-                        const serializedNode = parse5.serialize(fragment, serializerOpts);
+                    assertLocation(location, serializedNode, html, lines);
 
-                        assertNodeLocation(location, serializedNode, html, lines);
-
-                        // TODO: None of the cases below are ever matched.
```
fb55, line 140: This is the TODO that started this PR.

test/utils/generate-location-info-parser-tests.ts
```diff
@@ -113,42 +112,44 @@ export function generateLocationInfoParserTests(
             //Then for each node in the tree we run the serializer and compare results with the substring
             //obtained via the location info from the expected serialization results.
             it(`Location info (Parser) - ${test.name}`, async () => {
-                const serializerOpts = { treeAdapter };
-                const html = escapeString(test.data);
```
fb55, line 117: This is the `escapeString` call that invalidated the test.

wooorm (approved):

PR #442, opened 2022-03-07
fix(serializer): Serialize void elements as `""`

### LINKED ISSUE #289: Serialize the children of void elements as the empty string
The algorithm for [Serializing HTML fragments](https://html.spec.whatwg.org/multipage/parsing.html#serialising-html-fragments) was recently updated to serialize any potential children of void elements as the empty string, in https://github.com/whatwg/html/pull/4238.

fb55:
Fixes #289

### DIFF
--- a/packages/parse5/lib/serializer/index.test.ts
+++ b/packages/parse5/lib/serializer/index.test.ts
@@ -42,4 +42,15 @@ describe('serializer', () => {
 
         assert.equal(html, '<button>Hello</button>');
     });
+
+    it('serializes the children of void elements as the empty string (GH-289)', () => {
+        const document = parse5.parseFragment('<br/>');
+        const br = document.childNodes[0];
+        assert.ok(isElementNode(br));
+
+        // Add child node to `br`, to make sure they are skipped.
+        treeAdapters.default.appendChild(br, parse5.parseFragment('<div>').childNodes[0]);
+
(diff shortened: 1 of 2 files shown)


packages/parse5/lib/serializer/index.test.ts
```diff
@@ -42,4 +42,15 @@ describe('serializer', () => {
 
         assert.equal(html, '<button>Hello</button>');
     });
+
+    it('serializes the children of void elements as the empty string (GH-289)', () => {
+        const document = parse5.parseFragment('<br/>');
```
wooorm, line 47:
```suggestion
        const document = parse5.parseFragment('<br>');
```

IMO using the self-closing slash here gives off the idea that ***that*** is what makes an element “void”. It’s not, the slash is an XML-ism that does not do anything in HTML. Void elements are elements with a certain known tag name.

wooorm (commented):
One nit, otherwise :+1:

packages/parse5/lib/serializer/index.test.ts
fb55: Makes sense, happy to update

wooorm:
An optional idea: it could be good to test what happens to these known “void” elements in SVG/MathML contexts. As there void from HTML does not exist, but self-closing from XML does.

PR #447, opened 2022-03-11
test: Add upstream html5lib-tests

fb55:
And use them for the tokenizer, benchmarks & to generate feedback tests.

https://github.com/html5lib/html5lib-tests/pull/144 got merged, which allows us to adopt the upstream tests.

### DIFF
--- a/packages/parse5/lib/tokenizer/index.test.ts
+++ b/packages/parse5/lib/tokenizer/index.test.ts
@@ -1,4 +1,3 @@
-import * as parse5 from 'parse5';
 import { Tokenizer } from 'parse5/dist/tokenizer/index.js';
 import { generateTokenizationTests } from 'parse5-test-utils/utils/generate-tokenization-tests.js';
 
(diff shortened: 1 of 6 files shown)


packages/parse5/lib/tokenizer/index.test.ts
```diff
@@ -8,10 +7,3 @@ const tokenizerOpts = {
 };
 
 generateTokenizationTests('Tokenizer', dataPath.pathname, (handler) => new Tokenizer(tokenizerOpts, handler));
-
-describe('tokenizer', () => {
```
fb55, line 12: This test was upstreamed in https://github.com/html5lib/html5lib-tests/pull/145

test/utils/generate-parsing-tests.ts
```diff
@@ -118,7 +118,8 @@ function createParsingTest<T extends TreeAdapterTypeMap>(
     };
 }
 
-const treePath = new URL('../data/html5lib-tests/tree-construction', import.meta.url);
+// TODO: Stop using the fork here.
```
43081j, line 121: do we have an issue somewhere for this? would be good to mention the number i think

43081j (approved):

test/utils/generate-parsing-tests.ts
fb55: Opened as https://github.com/inikulin/parse5/issues/448

PR #449, opened 2022-03-13
chore(test): Make it possible to ignore tests

fb55:
And start running the upstream parser tests.

Also removes the upstreamed regression test (https://github.com/html5lib/html5lib-tests/pull/145)


test/data/tree-construction-regression/gh40_form_in_template.dat
```diff
@@ -1,13 +0,0 @@
-#data
```
wooorm, line 1: What’s the reason this has to be removed? Is it wrong?

fb55: It was upstreamed in https://github.com/html5lib/html5lib-tests/pull/145. I didn't remove it before because we didn't actually run the upstream html5lib-tests tree construction tests. Now that we do, it is save to remove it.

wooorm: ah okay, makes sense!

wooorm (approved):

fb55:
@wooorm Thanks for the review!

PR #450, opened 2022-03-13
chore: Remove `eval` from `scripting.test.ts`

fb55:
We had to fight with eslint & TS to make this happen. Now, we use Node's `vm` module instead.


43081j (approved):
Much cleaner 🎉

PR #453, opened 2022-03-13
fix(package): Allow deep imports from `parse5`

fb55:
The `generate-parser-feedback-test` script currently fails due to deep imports from `parse5` not being allowed. I assume this will apply to all submodules that currently have deep imports.

For now, let's allow deep imports of all files within `parse5`.

### DIFF
--- a/packages/parse5/package.json
+++ b/packages/parse5/package.json
@@ -33,8 +33,12 @@
     "types": "dist/index.d.ts",
     "exports": {
         ".": {
-            "import": "dist/index.js",
-            "require": "dist/cjs/index.js"
+            "import": "./dist/index.js",
+            "require": "./dist/cjs/index.js"
+        },
+        "./dist/*": {
+            "import": "./dist/*",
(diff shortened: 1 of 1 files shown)


43081j (approved):

packages/parse5/package.json
```diff
@@ -33,8 +33,12 @@
...
     "exports": {
         ".": {
-            "import": "dist/index.js",
-            "require": "dist/cjs/index.js"
+            "import": "./dist/index.js",
+            "require": "./dist/cjs/index.js"
+        },
+        "./dist/*": {
+            "import": "./dist/*",
+            "require": "./dist/cjs/*"
```
wooorm, line 41: I think that this is an anti-pattern.
The export map allows for an “allow-list” of what APIs to export as separate entry points, that is *unrelated* to the file system.
I believe we should carefully choose to expose what we want folks to use instead of exposing the whole file system?

Perhaps the problem in that script can be solved in a different way?

PR #455, opened 2022-03-13
chore: set non-null-assertion rule to error

### LINKED ISSUE #377: build improvement: disallow non-null assertions via ESLint

43081j:
Depends on #454 , draft until then.

Fixes #377

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -36,7 +36,7 @@
                 "prettier"
             ],
             "rules": {
-                "@typescript-eslint/no-non-null-assertion": "warn",
+                "@typescript-eslint/no-non-null-assertion": "error",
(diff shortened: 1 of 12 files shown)


43081j:
closing since actually we do need to use this in some places (especially dealing with stacks)

PR #461, opened 2022-03-16
refactor: Remove `getNextToken` method

fb55:
Removes the `getNextToken` function of the tokenizer, as well as all parsing loops. `getNextToken` was already a bit of a misnomer, as it would emit tokens to token handlers instead of returning it to the caller. Turns out parsing loops aren't necessary anymore, so let's get rid of them.

Script handling had to be updated slightly to support this. Now, a script handler is supplied to the parser constructor, and will be called whenever a script is encountered. This simplifies script handling logic in a few places.


wooorm (approved):
lgtm, I don’t see anything weird :)

PR #462, opened 2022-03-16
chore(test): Use `finished` to check streams

fb55:
The `finish` event doesn't actually fire every time a stream ends. This was already discovered in #432. This PR cleans up all other call locations.

Also switches us to using promises instead of callbacks for these tests. This has the advantage that errors will actually be reported by the test runner, instead of being uncaught inside a hanging test.


wooorm (commented):
I’m abstaining as I don’t feel qualified to review this :)

fb55:
> I’m abstaining as I don’t feel qualified to review this :)

No worries — as this only relates to tests, I'm feeling quite confident in the changes and will go ahead with things.

PR #467, opened 2022-03-18
chore(docs): Increase Node memory limit for docs

fb55:
`typedoc` builds successfully on my local ARM Mac, but runs out of memory on GitHub Actions. I suspect that the limit needs to be bumped slightly to make things go for now.

This is a known issue: https://github.com/TypeStrong/typedoc/issues/1606

This PR also includes a config change of typedoc to stop emitting source code, and to exclude functions marked as `@internal`. Plus, ESLint will now ignore built docs.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -21,6 +21,8 @@
         "out": "docs/build",
         "name": "parse5",
         "readme": "README.md",
+        "emit": "docs",
+        "excludeInternal": true,
         "entryPointStrategy": "packages"
     }
 }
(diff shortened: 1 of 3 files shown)


wooorm (approved):

tsconfig.json
```diff
@@ -21,6 +21,8 @@
         "out": "docs/build",
         "name": "parse5",
         "readme": "README.md",
+        "emit": "docs",
```
Gerrit0, line 24: Note: This is the default

fb55: Ah, I didn't realise — thanks for flagging.

This PR fixed our builds by the way, as can be seen in https://github.com/inikulin/parse5/commits/gh-pages. One way I was able to replicate the issue locally is to reduce the memory limit of Node using `--max-old-space-size=1024`.

PR #471, opened 2022-03-24
refactor(parse5-html-rewriting-stream): Move docs

fb55:
Moved docs to an interface, using the same pattern as `SAXParser`.


43081j (approved):
Looks good

wooorm (approved):

PR #476, opened 2022-03-24
chore(test): Update parser feedback tests

fb55:
Now that we use the upstream html5lib tests, we should update the parser feedback tests. I looked into adding support for fragments as well, but all tests pass in their current state 🙄 — let's table this for another time. The existing tests have been added.

There are 5 tests for oddities of the `<plaintext>` tag that break right now, and haven't been included.


wooorm (commented):
Every change in `test/` seems to be the same, but is so big I can’t review it.
`scripts/` is fine.

fb55:
> Every change in `test/` seems to be the same, but is so big I can’t review it.

All of them are generated, so this should be okay as long as `scripts` is alright.

PR #484, opened 2022-03-30
refactor: Use objects for adapters

fb55:
@43081j already had this in his TS conversion. Makes us use objects for the adapters, which allows TS to properly type check them. This means we will have to type the default options as `{Parser,Serializer}Options<any>`; please let me know if you can think of an alternative approach to make this type safe.

### DIFF
--- a/packages/parse5/lib/parser/index.ts
+++ b/packages/parse5/lib/parser/index.ts
@@ -1,7 +1,7 @@
 import { TokenHandler, Tokenizer, TokenizerMode } from '../tokenizer/index.js';
 import { OpenElementStack, StackHandler } from './open-element-stack.js';
 import { FormattingElementList, ElementEntry, EntryType } from './formatting-element-list.js';
-import * as defaultTreeAdapter from '../tree-adapters/default.js';
+import { defaultTreeAdapter } from '../tree-adapters/default.js';
 import * as doctype from '../common/doctype.js';
 import * as foreignContent from '../common/foreign-content.js';
 import { ERR, ParserErrorHandler } from '../common/error-codes.js';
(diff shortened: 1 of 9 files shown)


43081j:
awesome, this'll make the type system much happier :D 

i'll have a read through

packages/parse5/lib/parser/index.ts
```diff
@@ -111,7 +111,7 @@ export interface ParserOptions<T extends TreeAdapterTypeMap> {
     onParseError?: ParserErrorHandler | null;
 }
 
-const defaultParserOptions = {
+const defaultParserOptions: Required<ParserOptions<any>> = {
```
43081j, line 114: you could probably type this as the default adapter's map rather than `any`.

when you come to using it in the constructor/bootstrap, though, you'll have to cast it. IMO thats better than this one being untyped though.

whatever types we come up with for tree adapters, this'll always be the case i think. the only way around would be to have some common interface _without_ `treeAdapter` and fallback the adapter separately so it doesn't interfere with the types (you'd still need a cast i think).

packages/parse5/lib/serializer/index.ts
```diff
@@ -63,7 +63,7 @@ export interface SerializerOptions<T extends TreeAdapterTypeMap> {
 
 type InternalOptions<T extends TreeAdapterTypeMap> = Required<SerializerOptions<T>>;
 
-const defaultOpts = { treeAdapter: DefaultTreeAdapter, scriptingEnabled: true };
+const defaultOpts: InternalOptions<any> = { treeAdapter: defaultTreeAdapter, scriptingEnabled: true };
```
43081j, line 66: this one should probably be the type of the default adapter too

packages/parse5/lib/parser/index.ts
fb55: Made that change 👍

43081j (approved):
Looks good to me

Issue #301, opened 2019-10-02
Expose default treeadapter


jacobbogers:
I can now do:
```javascript
const ta = require('parse5/lib/tree-adapters/default');
```
but aesthetically not gorgeous like the rest of this wonderful code and  a bit of a hacky , could the default parser be exported as `require('parse5/adaptor-default')` or `require(`parse5/default-adaptor`);` or `const { adapter } = require('parse5');` and expose it officially this way

My 2cts

frank-dspeed:
@jacobbogers  vote 4 my issue #307  :) i will refactor to clean esm and produce babel cjs compatible version

jacobbogers:
I think your proposal covers refactoring to ES2020 standard, but I just want expose  "hidden" treeadapter, I will make PR for this, no worries, about your issue, lets discuss in your issue.

frank-dspeed:
@jacobbogers i will refactor to ES2020 and compile down to what ever the consumer needs and that will expose everything as a module including the treeAdapter :) 

```
import { treeAdapter } from 'es-parse5' 
```

PR #485, opened 2022-04-01
refactor: Be stricter about imports

fb55:
Addresses https://github.com/inikulin/parse5/pull/453#discussion_r828990730. Changes are in three categories:

1. Adds exports to the main `parse5` package (many of them marked as `@internal`, which hides them in the generated documentation). This way, we won't need deep exports. I'm not entirely happy with the amount of API surface this exposes, but marking these as `@internal` should allow us to introduce some potentially breaking changes in non-major versions.
2. Moves some code that made more sense in another place. This is for serialising doctypes (only used in the htmlparser2-adapter, as well as some tests), and `hasUnescapedText`, which made sense as part of the HTML file.
3. Standardises imports. This means we don't `import * as parse5` in most places, and we are explicit with type imports.

Additionally, all `package.json` files have been updated.

Fixes https://github.com/inikulin/parse5/issues/301


packages/parse5/lib/common/html.ts
```diff
@@ -1,4 +1,5 @@
-export enum NAMESPACES {
+/** All valid namespaces in HTML. */
+export enum NS {
```
fb55, line 2: We always aliased `NAMESPACES` as `NS`, even in this file. This makes it the actual name.

packages/parse5-html-rewriting-stream/package.json
```diff
@@ -18,7 +18,9 @@
     ],
     "license": "MIT",
     "main": "dist/index.js",
-    "exports": "dist/index.js",
+    "module": "dist/index.js",
+    "types": "dist/index.d.ts",
```
43081j, line 22: by setting this, does that mean importing an exact file will fail in typescript? since it'd look for the definition in here rather than `{whatever-you-imported}.d.ts`. maybe thats your intention? still trying to understand the overall change tbh so could've just misunderstood

fb55: With this PR, exact file imports are no longer allowed — all you can do is import the modules. This is done through the `exports` field.

As for the `types` field — from my understanding, this is for TS to figure out how to resolve a plain module import (ie. `import ... from 'parse5'`). TS should use co-located `.d.ts` files when you actually import a file.

43081j: because we don't bundle anything, i feel like we don't need it. typescript will already detect that there's an `index.d.ts` for the `index.js`. using a bare import will resolve `main` or whatever else, and then typescript will find the parallel `d.ts` file iirc.

fb55: I am honestly not sure if we need it or not. Eg. `tsc` might be able to pick up the type definitions, but will NPM show the TS badge if there is no `types` field in the `package.json`?

My gut instinct is to keep this field, just to be safe.

43081j: i just tried it on a package, with `types` set using a direct import of a module. it seems to work fine, so i think you're ok to leave this here

Issue #448, opened 2022-03-12
Complete switch to upstream html5lib tests


fb55:
A first step was completed in #447. We still rely on the fork for parser tests.

43081j:
@fb55 it looks like the fork just introduces `missing-doctype` and `non-conforming-doctype`

are you suggesting we keep those errors but somehow patch the html5lib tests to include them?

PR #487, opened 2022-04-01
feat(parser-stream): Support parsing fragments

### LINKED ISSUE #227: Implement ParserStream for HTML fragments
ParserStream ([docs](http://inikulin.github.io/parse5/classes/parserstream.html), [code](https://github.com/inikulin/parse5/blob/c7f11937701da4505d2420dcfe1cae50b57b9c41/lib/parser/parser_stream.js)) does not currently support parsing HTML fragments.
At least that's the impression I got looking at the docs - which don't mention it - and at the code, which is not nearly as complex as [parseFragment](https://github.com/inikulin/parse5/blob/c7f11937701da4505d2420dcfe1cae50b57b9c41/lib/parser/index.js#L327).

The use case would be to allow for a streaming load method in Cheerio: https://github.com/cheeriojs/cheerio/issues/1126#issuecomment-353882952

What do you think?

fb55:
Fixes #227

### DIFF
--- a/packages/parse5-parser-stream/test/parser-stream.test.ts
+++ b/packages/parse5-parser-stream/test/parser-stream.test.ts
@@ -4,7 +4,23 @@ import { generateParsingTests } from 'parse5-test-utils/utils/generate-parsing-t
 import { parseChunked } from './utils/parse-chunked.js';
 import { finished } from 'parse5-test-utils/utils/common.js';
 
-generateParsingTests('ParserStream', 'ParserStream', {}, (test, opts) => parseChunked(test, opts));
+generateParsingTests(
+    'ParserStream',
+    'ParserStream',
+    {
+        expectErrors: [
+            //NOTE: Foreign content behaviour was updated in the HTML spec.
+            //The old test suite still tests the old behaviour.
+            '269.foreign-fragment',
+            '270.foreign-fragment',
+            '307.foreign-fragment',
+            '309.foreign-fragment',
+            '316.foreign-fragment',
+            '317.foreign-fragment',
+        ],
+    },
+    (test, opts) => parseChunked(test, opts)
+);
 
 describe('ParserStream', () => {
     it('Fix empty stream parsing with ParserStream (GH-196)', async () => {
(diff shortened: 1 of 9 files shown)


packages/parse5-parser-stream/test/parser-stream.test.ts
```diff
@@ -4,8 +4,22 @@ import { generateParsingTests } from 'parse5-test-utils/utils/generate-parsing-t
...
 
-generateParsingTests('ParserStream', 'ParserStream', { skipFragments: true }, (test, opts) =>
-    parseChunked(test.input, opts)
+generateParsingTests(
+    'ParserStream',
+    'ParserStream',
+    {
+        expectErrors: [
+            //NOTE: Foreign content behaviour was updated in the HTML spec.
+            //The old test suite still tests the old behaviour.
```
wooorm, line 13: It’s a bit complex to read (for me) but it seems that p5 is correct, the tests (coming from another package) are wrong, which is why they are ignored.
Should there be an issue open(ed) there? And linked here? How long between HTML changes and the tests are updated? Do we need to help/fork? If its long, how do browsers handle this? 🤔

wooorm (approved):

packages/parse5-parser-stream/test/parser-stream.test.ts
fb55: We run the forked test suite here, not the upstream one. This is tracked in #448. The note here is copied from https://github.com/inikulin/parse5/blob/b428b7914483a8cef21a197bb2c9bb2aceb4527a/packages/parse5/lib/parser/index.test.ts#L16-L25

fb55: I have updated the comments to reference the issue.

wooorm: Oh right, I felt like I reviewed this already. Perhaps good to link that issue?

wooorm (approved):

fb55:
Thanks for the review @wooorm!

PR #496, opened 2022-04-10
feat(htmlparser2-adapter): Add CJS build

fb55:
And refactor how CJS is being built.

### DIFF
--- a/packages/parse5-htmlparser2-tree-adapter/package.json
+++ b/packages/parse5-htmlparser2-tree-adapter/package.json
@@ -14,13 +14,19 @@
         "htmlparser2"
     ],
     "license": "MIT",
-    "main": "dist/index.js",
+    "main": "dist/cjs/index.js",
     "module": "dist/index.js",
     "types": "dist/index.d.ts",
-    "exports": "./dist/index.js",
+    "exports": {
+        "import": "./dist/index.js",
+        "require": "./dist/cjs/index.js"
+    },
     "dependencies": {
         "domhandler": "^4.3.1",
         "parse5": "^6.0.1"
+    },
+    "scripts": {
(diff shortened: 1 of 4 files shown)


packages/parse5-htmlparser2-tree-adapter/package.json
```diff
@@ -14,14 +14,20 @@
         "htmlparser2"
     ],
     "license": "MIT",
-    "main": "dist/index.js",
+    "main": "dist/cjs/index.js",
```
43081j, line 17: are we sure the default entrypoint should be commonjs? shouldn't we be esm first and cjs second? i.e. only reference the cjs in the exports map

packages/parse5-htmlparser2-tree-adapter/package.json
```diff
@@ -14,14 +14,20 @@
...
     "module": "dist/index.js",
     "types": "dist/index.d.ts",
-    "exports": "./dist/index.js",
+    "exports": {
+        "import": "./dist/index.js",
```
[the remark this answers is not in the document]
fb55: This is a bit confusing, agreed. `build:esm` is in the root `package.json` (and make use of project references to build everything at once), the CJS ones are in the package's own `package.json`.

packages/parse5-htmlparser2-tree-adapter/package.json
fb55: Here is my thinking: The `main` field is only used by legacy NodeJS (everything else either uses the other fields). These versions don't have ESM support, and giving them anything but CJS wouldn't work.

43081j: yeah that makes sense. my concern was that tooling that also only looks at `main` would be forced to get CJS. though AFAIK most tooling these days does understand the `module` field or the `exports` field, so we're probably fine 👍

PR #510, opened 2022-04-20
chore: Add .npmignore

fb55:
#508 didn't work as expected; this adds an `.npmignore` file that excludes all unwanted files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,7 @@
 .vscode
 node_modules
 docs/build
-docs/05_api_reference.md
 packages/*/dist/
 test/dist/
-.DS_Store
 tsconfig.tsbuildinfo
+coverage/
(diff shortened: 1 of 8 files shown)


fb55:
I accidentally published a version of `parse5-html-rewriting-stream` with `.map` files — this also bumps the version of that package.

fb55:
Turns out this doesn't work either. Seems like ignores don't work with npm workspace publishes?

fb55:
I'm going with local file removals before publishing for now.

wooorm:
P.S. you can `npm pack` locally to see what gets packed up.

We should use `files` over an `.npmignore`: allow-lists are better at all this than ignore lists. Too frequently too much is published without a warning. But if important files are forgotten, well that stuff breaks so you get a notice to fix it.

wooorm:
Note: it’s one or the other.
And perhaps negated files don‘t work. Can you instead do `dist/**/*.js` instead if you want to include `.js` in `dist/`?

PR #516, opened 2022-04-20
chore(package): Make `files` a whitelist

fb55:
As proposed by @wooorm in https://github.com/inikulin/parse5/pull/510#issuecomment-1103825495


43081j (approved):
makes sense to me

wooorm (approved):

PR #517, opened 2022-04-20
refactor: Improve code coverage

fb55:
Removes all low-hanging fruit for #515

### DIFF
--- a/packages/parse5-htmlparser2-tree-adapter/lib/index.ts
+++ b/packages/parse5-htmlparser2-tree-adapter/lib/index.ts
@@ -149,9 +149,9 @@ export const adapter: TreeAdapter<Htmlparser2TreeAdapterMap> = {
             adapter.appendChild(document, doctypeNode);
         }
 
-        doctypeNode['x-name'] = name ?? undefined;
-        doctypeNode['x-publicId'] = publicId ?? undefined;
-        doctypeNode['x-systemId'] = systemId ?? undefined;
+        doctypeNode['x-name'] = name;
+        doctypeNode['x-publicId'] = publicId;
+        doctypeNode['x-systemId'] = systemId;
     },
 
     setDocumentMode(document: Document, mode: html.DOCUMENT_MODE): void {
--- a/packages/parse5-sax-parser/test/sax-parser.test.ts
+++ b/packages/parse5-sax-parser/test/sax-parser.test.ts
@@ -137,4 +137,21 @@ describe('SAX parser', () => {
 
         assert.throws(() => stream.write(buf), TypeError);
     });
+
+    it('Should treat NULL characters as normal text', async () => {
+        const parser = new SAXParser();
+        let foundText = false;
+
+        parser.on('text', ({ text }) => {
+            foundText = true;
+            assert.strictEqual(text, '\0');
+        });
+
+        parser.write('\0');
+        parser.end();
+
+        await finished(parser);
+
+        assert.ok(foundText);
+    });
 });
--- a/packages/parse5-sax-parser/test/sax-parser.test.ts
+++ b/packages/parse5-sax-parser/test/sax-parser.test.ts
@@ -152,6 +152,6 @@ describe('SAX parser', () => {
 
         await finished(parser);
 
-        assert.ok(foundText);
+        assert.strictEqual(foundText, true);
     });
 });
--- a/packages/parse5/lib/parser/open-element-stack.test.ts
+++ b/packages/parse5/lib/parser/open-element-stack.test.ts
@@ -317,6 +317,8 @@ generateTestsForEachTreeAdapter('open-element-stack', (treeAdapter) => {
     test('Has numbered header in scope', () => {
         const stack = new OpenElementStack(treeAdapter.createDocument(), treeAdapter, stackHandler);
 
+        assert.ok(stack.hasNumberedHeaderInScope());
+
         stack.push(createElement(TN.HTML), $.HTML);
         stack.push(createElement(TN.DIV), $.DIV);
         assert.ok(!stack.hasNumberedHeaderInScope());
@@ -337,6 +339,8 @@ generateTestsForEachTreeAdapter('open-element-stack', (treeAdapter) => {
     test('Has element in list item scope', () => {
         const stack = new OpenElementStack(treeAdapter.createDocument(), treeAdapter, stackHandler);
 
+        assert.ok(stack.hasInListItemScope($.P));
+
         stack.push(createElement(TN.HTML), $.HTML);
         stack.push(createElement(TN.DIV), $.DIV);
         assert.ok(!stack.hasInListItemScope($.P));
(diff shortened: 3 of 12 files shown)


packages/parse5/lib/tokenizer/index.ts
```diff
@@ -3050,20 +3052,6 @@ export class Tokenizer {
         }
     }
 
-    // Decimal character reference start state
-    //------------------------------------------------------------------
-    private _stateDecimalCharacterReferenceStart(cp: number): void {
```
fb55, line 3055: The _Decimal character reference start state_ is called in the _Numeric character reference state_ and immediately transitions to another state. Because of that, it will never be called in `_callState`, so let's inline it.

packages/parse5-htmlparser2-tree-adapter/lib/index.ts
```diff
@@ -149,9 +149,9 @@ export const adapter: TreeAdapter<Htmlparser2TreeAdapterMap> = {
             adapter.appendChild(document, doctypeNode);
         }
 
-        doctypeNode['x-name'] = name ?? undefined;
```
fb55, line 152: These values aren't nullable.

packages/parse5/lib/parser/open-element-stack.test.ts
```diff
@@ -317,6 +317,8 @@ generateTestsForEachTreeAdapter('open-element-stack', (treeAdapter) => {
     test('Has numbered header in scope', () => {
         const stack = new OpenElementStack(treeAdapter.createDocument(), treeAdapter, stackHandler);
 
+        assert.ok(stack.hasNumberedHeaderInScope());
```
fb55, line 320: The empty states won't be used in the parser, but we can test them here.

packages/parse5-sax-parser/test/sax-parser.test.ts
```diff
@@ -137,4 +137,21 @@ describe('SAX parser', () => {
...
+            foundText = true;
+            assert.strictEqual(text, '\0');
+        });
+
+        parser.write('\0');
+        parser.end();
+
+        await finished(parser);
+
+        assert.ok(foundText);
```
wooorm, line 155: probably a good idea to use an `assert.equal` instead, to check that the actual result makes sense and not, say, `'null'` for example?

packages/parse5/lib/parser/open-element-stack.test.ts
wooorm: 🤔 Why does `hasNumberedHeaderInScope` exist if parse5 doesn’t use it? Should we deprecate it and remove it in a major then?

wooorm: If these functions are used, can we instead add unit tests (input/output) of parsing behavior that depends on the result of these functions?

packages/parse5/lib/tokenizer/index.test.ts
```diff
@@ -1,9 +1,51 @@
...
+                assert.strictEqual(count++, 2);
+
+                expect(() => tokenizer.resume()).toThrow('Parser was already resumed');
+                tokenizer.write('<next>', true);
+            },
+            onStartTag(t): void {
+                assert.strictEqual(count++, 3);
+                assert.strictEqual(t.tagName, 'next');
+            },
+            onEndTag: noop,
```
wooorm, line 38: Why are all these required?
Should the default behavior instead become `noop`?

packages/parse5/lib/tokenizer/index.ts
```diff
@@ -132,7 +132,6 @@ const enum State {
     AMBIGUOUS_AMPERSAND,
     NUMERIC_CHARACTER_REFERENCE,
     HEXADEMICAL_CHARACTER_REFERENCE_START,
-    DECIMAL_CHARACTER_REFERENCE_START,
```
wooorm, line 135: I believe that this should either be upstreamed to the HTML spec or that it should be documented very clearly (in each line of removal) that this is a deviation from the spec and why. Even if the outcome is the same, the spec is complex, and people looking for bugs will notice that a state is missing.

packages/parse5/lib/tokenizer/index.ts
wooorm: See my comment above

packages/parse5/lib/tokenizer/index.ts
fb55: I've opened https://github.com/whatwg/html/pull/7841 to upstream this change.

packages/parse5/lib/parser/open-element-stack.test.ts
fb55: `hasNumberedHeaderInScope` is used by the parser. The only thing that the parser cannot reach is the default return value, eg. for an empty stack.

packages/parse5/lib/tokenizer/index.test.ts
fb55: The tokenizer callbacks are usually all needed. This test doesn't use a bunch of them as we tightly control the inputs.

packages/parse5/lib/parser/open-element-stack.test.ts
wooorm: You mean this early return https://github.com/inikulin/parse5/blob/8e22fe40d5774e2f2d5eb9be230f4f78d60e780d/packages/parse5/lib/parser/open-element-stack.ts#L270 causing https://github.com/inikulin/parse5/blob/8e22fe40d5774e2f2d5eb9be230f4f78d60e780d/packages/parse5/lib/parser/open-element-stack.ts#L278 to never run?

Could you change that early return to a break in that case, to fix the coverage?

Not sure about the algorithm: but it seems off that the default for that function i `true`? I’d expect, with such a function  name, to exit early with `true`, and otherwise return `false`?

packages/parse5/lib/tokenizer/index.ts
wooorm: sweet, thanks!

packages/parse5/lib/tokenizer/index.test.ts
wooorm: Your use of “usually” is rather “soft”. Does that mean you see something in making them optional?

packages/parse5/lib/parser/open-element-stack.test.ts
fb55: The relevant part of the spec is https://html.spec.whatwg.org/multipage/parsing.html#has-an-element-in-scope

The algorithm explicitly doesn't consider the possibility of an empty stack:

> This will never fail, since the loop will always terminate in the previous step if the top of the stack — an [html](https://html.spec.whatwg.org/multipage/semantics.html#the-html-element) element — is reached.

fb55: I have updated the logic to use `break`, but am still split if this is the right approach. Another option would be to error if the loop terminates.

packages/parse5/lib/tokenizer/index.test.ts
fb55: Sorry for not being clearer. This is the _only_ case where not all callbacks are needed. Every other time, we supply all of the relevant callbacks.

packages/parse5/lib/parser/open-element-stack.test.ts
wooorm: Then I recommend using a `c8 ignore` (or similar), with this as an explanation:

```
/* Note: the algorithm <link> never fails */
/* c8 ignore next 123 */
```

PR #518, opened 2022-04-21
docs: Fix small typo in code example

cheeaun:
Page: https://parse5.js.org/modules/parse5.html#parseFragment

Variable `parser` should be `parse5`. As it's declared in beginning of the code example `const parse5 = require('parse5');`, `parser` variable would be undefined.


fb55:
Good find, thanks!

PR #527, opened 2022-04-30
chore(deps): Bump domhandler, remove cast

fb55:
domhandler now allows `Document` nodes as children. This allows us to remove a cast from the relevant adapter.


fb55:
Not waiting for a review as (1) I don't want dependabot to beat me to this change and (2) none of the changes seem controversial in any way.

PR #631, opened 2022-07-28
feat(parse5): use string types for node names

43081j:
This was initially to change `NodeType` into a union, but that doesn't make sense since we wouldn't make use of the type then, we'd just use the string literal types anyway.

Nothing references `NodeType` outside the tree adapter and it isn't exported publicly (`index.ts`) currently.

So instead i just dropped `NodeType` and used a string literal type for each node type.

@fb55 what do you think? this matches how it was originally (pre-rewrite) and allows people to create nodes without pulling the (currently not exported afaict) enum in.

if anyone wants the type union, they can do `Node['nodeName']`, which we could alias into `NodeType` if its still of any use

### DIFF
--- a/packages/parse5/lib/tree-adapters/default.ts
+++ b/packages/parse5/lib/tree-adapters/default.ts
@@ -2,17 +2,9 @@ import { DOCUMENT_MODE, type NS } from '../common/html.js';
 import type { Attribute, Location, ElementLocation } from '../common/token.js';
 import type { TreeAdapter, TreeAdapterTypeMap } from './interface.js';
 
-export enum NodeType {
-    Document = '#document',
-    DocumentFragment = '#document-fragment',
-    Comment = '#comment',
-    Text = '#text',
-    DocumentType = '#documentType',
-}
-
 export interface Document {
     /** The name of the node. */
-    nodeName: NodeType.Document;
+    nodeName: '#document';
     /**
      * Document mode.
      *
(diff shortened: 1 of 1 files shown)


43081j:
@fb55 similarly, we can't get at `DOCUMENT_MODE` and `NS` for these cases:

```ts
const document: Document = {
  nodeName: '#document',
  mode: DOCUMENT_MODE.NO_QUIRKS,
  childNodes: []
};
```

`DOCUMENT_MODE` isn't exported in the root and doesn't have an export mapping as far as i can see.

same with `NS`.

exact same problem we had here with `NodeType` - it wasn't exported anywhere.

fb55 (approved):
Looks good to me!

43081j:
@fb55  any chance you could prep a release for this? im not sure what process you've been following so probably better you cut it

fb55:
I tried to publish a new release yesterday and didn't remember exactly what I did. Will try to figure things out next weekend, and document it for future releases.

fb55:
After pushing this in front of me for a while, I've now published `parse5@7.1.0`. Sorry for the delay!

The other modules didn't change, so no need to publish new versions for them. This also allowed me to not reproduce the `npm --workspaces` magic that I used the first time.

fb55:
Turns out that I did forget one thing, and that is opening a PR with the version bump _before_ publishing. Opened #662 with the change.

PR #640, opened 2022-08-12
chore(ci): Pin GitHub Action dependency versions

fb55:
Dependabot will now update them as new versions are released.

This is done to improve parse5's score in the OpenSSF scorecard, visible here: https://deps.dev/npm/parse5

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -20,14 +20,14 @@ jobs:
 
         steps:
             - name: Checkout repository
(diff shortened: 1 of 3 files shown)


43081j (approved):

wooorm:
I’d like to stress that bots are extremely noisy. Especially for patch releases. I routinely gets *tons* of notifications from GitHub, more than I can handle, bots are useless to me, so mark them as done. I’ll likely miss more important notifications due to this.

PR #662, opened 2022-09-02
Bump parse5 to 7.1.0

fb55:
This version is already on npm — forgot that I should PR the release first.

Once this is merged, I'll create a tag and publish release notes.

My gut feeling is that, going forward we should skip versions for the other packages, so that versions numbers across packages are monotonically increasing. (Ie. `parse5-rewriter-stream@7.8.9` is guaranteed to be published after `parse5@7.6.5`.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6012,7 +6012,7 @@
             }
         },
         "packages/parse5": {
-            "version": "7.0.0",
+            "version": "7.1.0",
             "license": "MIT",
             "dependencies": {
                 "entities": "^4.4.0"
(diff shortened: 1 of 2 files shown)


43081j (approved):

43081j:
if we do figure out a good workflow for it, could be nice to document it in the contributing file if we didn't already.

im guessing the order should be:

```bash
npm version patch
git push && git push -u origin x.y.z
npm publish
```

but not sure how it fits into workspaces etc either

fb55:
What I'd do going forward is: 

```
cd packages/parse5
npm version minor
```

Then open a PR with this change — this allows the release to be green-lit.

Once the PR is merged:

- Create a tag & release on GitHub
- `npm publish`

How does that sound?

43081j:
makes sense to me 👍

PR #715, opened 2022-10-27
fix(tokenizer): Properly handle CR when peeking

### LINKED ISSUE #710: CRLF inside `</noscript>` occurs `eof-in-element-that-can-contain-only-text` error
```html
<!DOCTYPE html>
<html>
  <body>
    <!-- prettier-ignore -->
    <noscript>foo</noscript
    >
  </body>
</html>
```

reproduction: https://stackblitz.com/edit/node-djhqfm?file=input-lf.html,index.js

If it's using LF instead of CRLF, the error won't happen. I'm not much familiar with HTML spec but I guess it should work in the same way since it's [normalized before the tokenization stage](https://html.spec.whatwg.org/#preprocessing-the-input-stream).

Original issue: https://github.com/vitejs/vite/issues/10580

fb55:
Peeking didn't properly support carriage returns, which led the parser to hang in special tags.

Fixes #710

### DIFF
--- a/packages/parse5/lib/parser/index.test.ts
+++ b/packages/parse5/lib/parser/index.test.ts
@@ -78,6 +78,13 @@ describe('parser', () => {
         expect(doctype).toHaveProperty('systemId', '');
     });
 
+    it('Regression - CRLF inside </noscript> (GH-710)', () => {
+        const onParseError = jest.fn();
+        parse('<!doctype html><noscript>foo</noscript\r\n>', { onParseError });
+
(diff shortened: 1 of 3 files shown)


wooorm (approved):
Looks good, but the test seems a bit specific to how the tokenizer is implemented. Perhaps a reduced case from the original problem (https://github.com/inikulin/parse5/issues/710#issue-1420861775) can be added as a test for the parser as a whole?

fb55:
Thanks for the review! I've switched the test to a parser regression test, following the pattern we already had in place for similar issues.

PR #716, opened 2022-10-30
feat(parser-stream): Add CJS build

fb55:
I want to use the parser stream in Cheerio, and am blocked by the current ESM nature of the module. The path of least resistance is to add a CJS build for the parser stream.


fb55:
Going ahead with this as it doesn't seem incredibly controversial, and I haven't heard anything in two weeks.

PR #874, opened 2023-04-01
Fixes exports used in the @parse5/tools types

arcanis:
The `@parse5/tools` package references `parse5/dist/tree-adapters/default.js` in its imports (for example [here](https://github.com/parse5/parse5-tools/blob/main/src/attributes.ts#L1)), but since the path isn't part of the `exports` field it fails to resolve under certain TS configurations.

### DIFF
--- a/packages/parse5/package.json
+++ b/packages/parse5/package.json
@@ -32,8 +32,13 @@
     "module": "dist/index.js",
     "types": "dist/index.d.ts",
     "exports": {
-        "import": "./dist/index.js",
-        "require": "./dist/cjs/index.js"
+        ".": {
+            "import": "./dist/index.js",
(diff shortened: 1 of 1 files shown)


wooorm (approved):

fb55:
The default tree adapter is exported by the main export. Not opposed to this PR, but it might be easier to update the import.

43081j:
from what i remember, it worked in the tools build thanks to it being a type-only import (which, at the time at least, didn't care about package exports).

if the default tree adapter's _types_ are available from the root, we should update p5 tools rather than the parse5 exports.

off the top of my head, im not sure if they are exposed. the adapter itself is, but unsure of the types

fb55:
The types have been exposed in #926.

PR #879, opened 2023-04-03
docs: fix variable name in sample code

nolanlawson:
This code sample is using the wrong variable name. The declared variable is `html`, but on line 104 it tries to log a variable called `str`.

In line with [the similar example in the same file](https://github.com/inikulin/parse5/blob/badd96506b99a4101fc71853720d689ca573d592/packages/parse5/lib/serializer/index.ts#L69-L72), I renamed the variable to `str`.


fb55:
Thanks @nolanlawson!

PR #890, opened 2023-04-10
fix: include sourcemap files in the dist

### LINKED ISSUE #831: Failed to parse source map ...
I am using parse5 in a react app, and the modules are referring to a sourceMap file that is not being found, which is interrupting the build of the app.

I can confirm checking the dist/index.js in npm

Line 45
```js
//# sourceMappingURL=index.js.map
```

however, there isn't such file relative to path

prashantpalikhe:
Fixes #831

### DIFF
--- a/packages/parse5-html-rewriting-stream/package.json
+++ b/packages/parse5-html-rewriting-stream/package.json
@@ -32,6 +32,7 @@
     },
     "files": [
         "dist/**/*.js",
+        "dist/**/*.js.map",
         "dist/**/*.d.ts"
     ]
 }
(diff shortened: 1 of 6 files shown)


fb55:
This will lead to the next issue, which is that source maps reference source files (which aren't part of the package). I've solved this in most of my modules by specifying [a `sourceRoot`](https://www.typescriptlang.org/tsconfig#sourceRoot) that points at the source files on GitHub. That won't work for parse5 however, as we are using `tsc --build` to build a number of packages at once, and `--build` does not allow `sourceRoot` to be set.

The best way forward here is probably to disable source maps during manual builds, or how that the TypeScript team will add support for `sourceRoot` in `--build`.

PR #897, opened 2023-04-13
refactor(tokenizer): Use entities' new decoder

fb55:
[`entities@4.5.0` features](https://github.com/fb55/entities/releases/tag/v4.5.0) an `EntityDecoder` that implements the entity decoding logic from this module in a way that can be shared between my three projects working with HTML entities.

With this change, the main use-case for hibernation is removed. The remaining use, waiting for sequences, can be removed in a future patch, which should make it easier to follow the logic in the tokenizer (I was surprised several times by unintuitive flows while working on this PR).

Ultimately, I'd like to see both htmlparser2 and parse5 become more modular, which will make them easier to maintain, and should provide a big speed boost — https://github.com/marko-js/htmljs-parser is currently the [fastest parser](https://github.com/AndreasMadsen/htmlparser-benchmark/blob/317d2db83f34e7dbd9c3ea05d432bfd740e0e42c/stats.txt) in https://github.com/AndreasMadsen/htmlparser-benchmark, using a modular design to great effect.


package.json
```diff
@@ -59,11 +59,6 @@
         "preset": "ts-jest/presets/default-esm",
         "testEnvironment": "node",
         "coverageProvider": "v8",
-        "globals": {
```
fb55, line 62: Unrelated change, but this removes a deprecation warning from `ts-jest`. There is no change in behavior.

fb55:
@wooorm @43081j I'd love your review if you are available!

wooorm:
Sorry, I have no knowledge of the `entities` package, or deep knowledge of the internals here, to judge this PR

43081j:
I'll have a read through 👍

43081j (approved):
sorry it took so long, i've been away recently

looks good to me, though i haven't had chance to read much into the entities package in your repo

fb55:
Thanks for the review!

PR #996, opened 2023-07-28
feat(parse5): change parser/tokenizer methods to be protected

### LINKED ISSUE #992: Migrate some or most methods of Tokenizer and Parser to `protected`
in the wild i've seen people inheriting `Parser` and `Tokenizer`, using 6.x.

upgrading to 7.x means, in typescript at least, they can no longer override methods and access internals.

we should probably just mark anything `private` as `protected` instead so they can at least be inherited and overridden while keeping them away from the public interface.

@fb55 thoughts?

43081j:
Some projects in the wild depend on extending `Parser` and `Tokenizer` by overriding certain methods and using internals we don't publish on our public API.

This just moves all the private members to be protected so those projects can upgrade to 7.x.

Not everything should be protected but it seems our current class members all fit the criteria.

Fixes #992 

i think this should be fine but let me know your thoughts @fb55

### DIFF
--- a/packages/parse5/lib/parser/index.ts
+++ b/packages/parse5/lib/parser/index.ts
@@ -123,7 +123,7 @@ export class Parser<T extends TreeAdapterTypeMap> implements TokenHandler, Stack
     treeAdapter: TreeAdapter<T>;
     /** @internal */
     onParseError: ParserErrorHandler | null;
-    private currentToken: Token | null = null;
+    protected currentToken: Token | null = null;
     public options: Required<ParserOptions<T>>;
     public document: T['document'];
 
@@ -230,7 +230,7 @@ export class Parser<T extends TreeAdapterTypeMap> implements TokenHandler, Stack
     /** @internal */
     activeFormattingElements: FormattingElementList<T>;
     /** Indicates that the current node is not an element in the HTML namespace */
-    private currentNotInHTML = false;
+    protected currentNotInHTML = false;
 
     /**
      * The template insertion mode stack is maintained from the left.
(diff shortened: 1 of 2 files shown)


43081j:
@fb55 any chance you can help with getting this over the line and releasing it?

it'll unblock some work in salesforce and google im trying to sort out

43081j:
@wooorm do you mean you're calling those private (now protected in this branch) methods from outside?

i.e. `parser._somePrivateMethod()`

as opposed to calling it within a subclass (`this._somePrivateMethod()`)

if thats the case, it suggests these are public methods really, yeah. maybe you could help with an example of why you need to call those methods publicly

wooorm:
See https://github.com/syntax-tree/hast-util-raw.
It’s a complex project that tightly integrates with how p5 works.
It uses the parser to pass *both* nodes through *and* strings of HTML, while keeping positional info and other data on objects around.
This is a particular problem that arises with markdown as you can both have nice nodes, and strings of HTML inside/around them.

43081j:
interesting!

i wonder if it'd make more sense that you implement that logic as a custom parser? you could inherit `Parser` to consume your custom data and call the exact same methods but from within your equivalent parse method.

something like that. 

but ofc thats a big job so its unlikely even if you agree, it'd get done any time soon i think. so im not sure... we can make them public so you don't need to refactor, or we can keep them protected and aim to one day rework your code

43081j:
@fb55  can you help me decide what to do here?

i don't want this to go stale as i have third party PRs waiting on it

looks like we need to make them `protected` or `public`, i don't mind which. `public` will help @wooorm out

43081j:
@wooorm are you ok if we go ahead with `protected` for now and do the larger public change later? since your code isn't blocked by it at least

i feel like there's probably a cleaner solution than just making everything public for you too, e.g. exposing more helpful public APIs you can use

wooorm:
I think it doesn’t break my code, indeed. So indeed, this does not affect me, and that’s okay!

43081j:
@fb55 not sure how you've been publishing releases, so would you mind getting one published some day? no rush

divmain:
I went to open an issue/PR asking for exactly this and discovered the exact change I was hoping for had just recently been merged. Kudos and thanks!

43081j:
@divmain thats because i also had a branch lying around for updating LWC's parse5! 😂 

i still have some other changes this will unblock in lwc though, so ill open a pr just for those

Issue #1031, opened 2023-09-11
Expose createTextNode() on default tree adapter


steffans:
Would be great if the createTextNode() function is also exposed on the default tree adapter to allow overrides an reuse logic of other tree adapter functions.

fb55:
PRs welcome!

43081j:
If anyone does tackle this, it'd be helpful to also open a PR in [parse5-tools](https://github.com/parse5/parse5-tools) since it currently provides its own `createTextNode` thanks to the lack of one here.

so you could remove that and just alias the new one you introduce here

fb55:
Closed in #1040

PR #1040, opened 2023-09-27
refactor: expose createTextNode on defaultTreeAdapter

jmsjtu:
This PR exposes the `createTextNode()` method on the default tree adapter (#1031).

I pulled down the tests from [`html5lib-tests`](https://github.com/html5lib/html5lib-tests/tree/master) and ran them locally but wasn't sure there's any additional unit testing that should be included as part of this change.

I'm planning to open a follow-up PR on [parse5-tools](https://github.com/parse5/parse5-tools) to replace the existing `createTextNode` with the one provided here as described in #1031.

### DIFF
--- a/packages/parse5-htmlparser2-tree-adapter/lib/index.ts
+++ b/packages/parse5-htmlparser2-tree-adapter/lib/index.ts
@@ -27,10 +27,6 @@ export type Htmlparser2TreeAdapterMap = TreeAdapterTypeMap<
     ProcessingInstruction
 >;
 
-function createTextNode(value: string): Text {
-    return new Text(value);
-}
-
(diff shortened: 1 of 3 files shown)


43081j:
looks good to me 👍 

as far as tests go - im not sure either if we explicitly test the adapter methods anywhere. i imagine it indirectly gets tested by the other, wider tests. so maybe its already covered

43081j:
@fb55  could you take a look at this too? if you could help me publish a new version after we merge this, that'd be much appreciated also. we get this and the `protected` stuff i did a while ago, then

fb55:
Thanks @jmsjtu!

PR #1113, opened 2024-01-25
fix: don't generate sourcemap on build, close #831

### LINKED ISSUE #831: Failed to parse source map ...

antfu:
I am not sure if sourcemaps are used somewhere, but since it's not shipped with the npm, I think it's better to not generate them otherwise would lead to hanging sourcemap reference #831

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,6 @@
                 "lint-staged": "^15.2.0",
                 "outdent": "^0.8.0",
                 "prettier": "^3.2.2",
-                "rimraf": "^5.0.5",
                 "ts-jest": "^29.1.1",
                 "ts-node": "^10.9.2",
                 "typedoc": "^0.25.7",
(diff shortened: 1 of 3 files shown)


43081j:
makes sense to me but have added felix as a reviewer as he might know more

wooorm:
The code here might also be because it was intended to publish declaration/source maps, right? Could solving that by adding it to “files” also solve your issue?

antfu:
The previous PR trying that way gets closed:
https://github.com/inikulin/parse5/pull/890#issuecomment-1503282366

I don't mind either ways - it's just the current state causing build tools like Vite emits a lot of warnings

fb55:
Thanks @antfu!

JvanderHeide:
Could we get a release for this?

PR #1266, opened 2024-09-19
chore: migrate to mocha and eslint 9

43081j:
Migrates from Jest to mocha and node's own `assert` library.

Also updates ESLint to 9.x and bumps TSESLint to do the same.

@fb55 what do you think of this? jest is pretty hefty for what we need, this trims a fair chunk of the dependency weight out. mocha, tinyspy and node `assert` seem pretty clean together to me

similarly, I dropped most eslint customisation we had and just opted to rely on the recommended configs since prettier should pick up all the stylistic stuff we care about


packages/parse5-htmlparser2-tree-adapter/package.json
```diff
@@ -26,7 +26,7 @@
         "parse5": "^7.0.0"
     },
     "scripts": {
-        "build:cjs": "tsc --module CommonJS --target ES6 --outDir dist/cjs && echo '{\"type\":\"commonjs\"}' > dist/cjs/package.json"
+        "build:cjs": "tsc --moduleResolution node10 --module CommonJS --target ES6 --outDir dist/cjs && echo '{\"type\":\"commonjs\"}' > dist/cjs/package.json"
```
43081j, line 29: typescript started to complain without this, since you _must_ have `module: "node16"` if you have module resolution set to it (which we do)

I guess it got updated in the lockfile while updating tseslint etc and introduced the new behaviour

fb55:
Nice!

Could we use Node's own test runner and skip having any dependencies?

43081j:
we could but I wasn't sure what node version matrix we wanted

the node test runner is available in 16.x onwards if I remember correctly

43081j:
maybe lets go with mocha for now and I can try the node runner in a separate PR? since I feel like it'll be a bit more work

wooorm:
I’d love all this stuff, but separate PRs makes reviewing this easier!

43081j:
there's basically two changes (mocha and eslint). I could open two separate PRs but it gets a bit weird since I need to setup eslint to be happy with jest, knowing it'll go away immediately after

it may turn out its as easy as `globals.jest` but I'll have to try it out to find out

if you're unhappy with it going ahead without me splitting it up though I can have a look

wooorm:
Right, sure. There are ways around that, with things like `// todo(@43081j): remove this workaround when using builtin node test` or so.

I’m not going to block this. But in my experience splitting PRs up into different things helps.

43081j:
yes, I think we should go ahead with mocha first though anyway since the node test runner will be more work

I'll see if I can split into two PRs

it wouldn't be a TODO for removing a workaround for `node:test`, it'd be temporary lint config to make ESLint happy about jest globals and what not. we'd then remove that in the mocha PR

43081j:
Closing in favour of #1268 and #1267

PR #1267, opened 2024-09-19
chore: upgrade to eslint 9.x

43081j:
Upgrades to eslint 9.x and simplifies the customisation of our config.


bench/perf/index.js
```diff
@@ -7,6 +7,8 @@ import * as parse5 from '../../packages/parse5/dist/index.js';
 import { ParserStream as parse5Stream } from '../../packages/parse5-parser-stream/dist/index.js';
 import * as parse5Upstream from 'parse5';
 
+/* eslint-disable no-console */
```
wooorm, line 10: I don’t see a good reason to put file based ignores after import (export?) and before global definitions. I’d say before the `import`s makes more sense?

eslint.config.js
```diff
@@ -0,0 +1,69 @@
...
+            '@typescript-eslint/no-non-null-assertion': 'warn',
+            '@typescript-eslint/no-explicit-any': 'warn',
+            '@typescript-eslint/explicit-function-return-type': 'error',
+            '@typescript-eslint/consistent-type-imports': 'error',
+
+            '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
+        },
+    },
+    eslintConfigPrettier,
+];
```
wooorm, line 69: This looks like it is removing `plugin:unicorn/recommended`? Why?

packages/parse5-htmlparser2-tree-adapter/package.json
```diff
@@ -26,7 +26,7 @@
         "parse5": "^7.0.0"
     },
     "scripts": {
-        "build:cjs": "tsc --module CommonJS --target ES6 --outDir dist/cjs && echo '{\"type\":\"commonjs\"}' > dist/cjs/package.json"
+        "build:cjs": "tsc --moduleResolution node10 --module CommonJS --target ES6 --outDir dist/cjs && echo '{\"type\":\"commonjs\"}' > dist/cjs/package.json"
```
wooorm, line 29: > 'node10' (previously called 'node') for Node.js versions older than v10, which only support CommonJS require. You probably won’t need to use node10 in modern code.

a) why is such an old thing needed here,
b) what does it have to do with eslint?

bench/perf/index.js
43081j: Sure I just threw it somewhere near the top

Happy to move it

eslint.config.js
43081j: It is

Much of it is covered by the formatter or the recommended rule sets

It would be good to know what exactly we wanted it to catch. It's a very big plugin for not much benefit by the looks of it

packages/parse5-htmlparser2-tree-adapter/package.json
43081j: Typescript will refuse to build with node16 resolution and commonjs module

Our config was wrong all along but the older tsc just didn't complain

Presumably installing tseslint bumped it in the lock file

So either we do this or we fix cjs builds properly (which means we can't just run a build twice but need a separate config which doesn't have type: module or vice versa)

eslint.config.js
wooorm: Re the first paragraph, much of ESlint can also be done by hand, yet we still use ESlint.

Re the second, that is very subjective, and I strongly disagree.
Removing lint rules is always silent. The project was formatted. So of course you don‘t notice that lint rules are useful.

Re the last paragraph, I even stronger disagree. Lint rules are very useful in my experience.

packages/parse5-htmlparser2-tree-adapter/package.json
fb55: I switched my packages to https://github.com/isaacs/tshy to work around this (and improve the source maps situation).

eslint.config.js
43081j: The last paragraph was very far from suggesting lint rules are not useful. We pulled a very large package in without really knowing which rules we cared about other than the ones we manually enabled (i.e I don't think any of us ever knew what was in the recommended config) 

Much of it is stylistic and should be owned by the formatter (and already is). But if there are other rules in there we think are useful, I can reintroduce it

The recommended configs have grown since and cover all the more important stuff.

wooorm: @fb55 has intentionally pulled these rules in. I agree with these rules. Many rules are not things prettier does: https://github.com/sindresorhus/eslint-plugin-unicorn#rules. This change has nothing to do with the PR title.

43081j: The PR removes many of our customisations, so it certainly was relevant

I don't care enough either way so I will reintroduce it some day when I get time and move past this

Once a more focused eslint package exists, maybe we can drop it for good. Overly generic, unfocused plugins are not great

packages/parse5-htmlparser2-tree-adapter/package.json
43081j: Id rather not do that in this pr, especially given @wooorm raising that we need to split PRs up

So maybe let's move ahead with this since the config was wrong before anyway? And I'll open a separate pr some day to use tshy

eslint.config.js
wooorm: This PR is titled “upgrade to eslint”, not “remove things I find irrelevant”.
I do care.
I don’t think you are correct in calling things you dislike unfocussed or overly generic.
I am taken aback that you want to push things through that don’t have consensus.

@fb55 Please chime in.

How is this project governed?

fb55: Just because unicorn is broad does not mean it doesn't have value 😄

I am generally a fan of [Isaac Schlueter's Principle of Most Restrictive Production](https://blog.izs.me/2022/04/principle-of-most-restrictive-production) and enjoy when lint tools push a codebase in that direction. Unicorn has a number of rules that help with that, eg. for regex, or switch statements.

Most of the interesting unicorn rules have been ported to Biome, so that might be an alternative we could migrate to.

@wooorm Let's not dig in to each other too much, this is how maintainers burn out 😅

43081j: The original post specifies that I simplified the customisations we have. It was part of that. I would have removed some of the custom rule options too but it turns out they're not in the recommended rule set anymore. 

It was part of the described change regardless of the title. Apologies for not having a longer title

Too much energy has been lost discussing this as I don't care enough about removing it to take part in a debate

I will very simply reintroduce it 👍

wooorm: @fb55 I would still appreciate info on how this project is governed. Consensus-seeking? Vote? Consensus with a vote as backup? Dictator?
Was there something agreed with Ivan?
As you mention maintainer burnout, and you once asked me to help, and there is no `governance.md`, I’d like to hear your thoughts.

fb55: No agreement with Ivan. The three of us are responsible for the project, and should try to move it forward in good faith. There has been little activity in the repo for the past year, and setting up a proper governance structure does not seem like an immediate need.

Super happy to see @43081j push things forward, and I do believe that we should support these efforts instead of fighting for small stuff.

packages/parse5-htmlparser2-tree-adapter/package.json
fb55: Sounds good!

bench/perf/index.js
43081j: turns out it was like that before and I had just left it as-is

I've moved it now, though

43081j:
@wooorm and/or @fb55 can you take another look and I'll get this one merged so we can get the mocha one sorted too

fb55 (approved):
Looks great! Thanks for doing this

wooorm (commented):
looks good!

PR #1268, opened 2024-09-19
chore: migrate from jest to mocha

43081j:
Migrates from jest to mocha, c8 and tinyspy.

### DIFF
--- a/.c8rc.json
+++ b/.c8rc.json
@@ -1,5 +1,5 @@
 {
     "reporter": ["lcov"],
     "include": ["packages/**/*.ts"],
-    "exclude": ["**/*.test.ts"]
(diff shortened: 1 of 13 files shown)


43081j:
The coverage drop suggests we missed some test files. I'll take a look 👍

Also this depends on #1267 to avoid having to do the lint config twice

43081j:
the exact same number of tests run but the coverage is way down on the tokenizer, so I'm taking a look into that 

super strange since the coverage is clearly setup to find the right files, it just calculates a much lower value

maybe jest was lying 🤔

43081j:
giving up on this for #1376 instead which has working coverage

PR #1277, opened 2024-09-30
fix: fix `<noframes>` parsing

### LINKED ISSUE #972: Noframes tag not parsed according to specification when in body
Hello,

I found that parse5 does not adhere to the HTML specification when it comes to parsing the `noframes` tag outside of the `head` of a document.

[According to the specification](https://html.spec.whatwg.org/multipage/parsing.html#parsing-main-inbody), when a `noframes` tag is encountered inside a document's `body`, it should be processed as if it was in `head`, i.e., with raw text content. Parse5 however seems to parse it as if it were a custom tag. For the other tags listed in the spec it behaves correctly, but `noframes` seems to be missing [here](https://github.com/inikulin/parse5/blob/master/packages/parse5/lib/parser/index.ts#L2383)

Example:
```html
<r><noframes><math id="</noframes><b>should be outside of noframes</b>">
```
[Explorer link](https://astexplorer.net/#/gist/becbddd607b94ab613750feee1c46ecf/374e85b2445b81300727601b528d531d61b128f4).

In `head` mode the content of `noframes` should be parsed as raw text, that is the closing tag inside the id attribute of the `math` tag should close the opening `noframes` tag and the following `b` tag should be outside of `noframes`.

nolanlawson:
Fixes #972

This also adds a test to make sure that "raw text" tags (e.g. `<noembed>`, `<noframes>`, `<style>`, etc.) all work the same. For the test, I'm reusing the same repro from #972.

Before this fix, all the tests pass except for `<noframes>` specifically.

### DIFF
--- a/packages/parse5/lib/parser/index.test.ts
+++ b/packages/parse5/lib/parser/index.test.ts
@@ -3,6 +3,7 @@ import { parseFragment, parse } from 'parse5';
 import { jest } from '@jest/globals';
 import { generateParsingTests } from 'parse5-test-utils/utils/generate-parsing-tests.js';
 import { treeAdapters } from 'parse5-test-utils/utils/common.js';
+import type { Element, TextNode } from '../tree-adapters/default.js';
 
 generateParsingTests(
     'parser',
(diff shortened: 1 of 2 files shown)


packages/parse5/lib/parser/index.test.ts
```diff
@@ -110,4 +111,41 @@ describe('parser', () => {
...
+            expect(target.childNodes).toHaveLength(1);
+            expect(target.childNodes[0].nodeName).toBe('#text');
+            expect((target.childNodes[0] as TextNode).value).toBe('<math id="');
+
+            const b = r.childNodes[1] as Element;
+            expect(b.childNodes).toHaveLength(1);
+            expect(b.childNodes[0].nodeName).toBe('#text');
+            expect((b.childNodes[0] as TextNode).value).toBe('should be outside');
+        });
+    });
```
nolanlawson, line 150: I couldn't find a better place to put this test. For some reason the tests in `test/data/html5lib-tests/serializer` do not repro the issue.

packages/parse5/lib/parser/index.test.ts
```diff
@@ -110,4 +111,41 @@ describe('parser', () => {
...
+        it.each([
+            ['iframe'],
+            ['noembed'],
+            ['noframes'],
+            ['noscript'],
+            ['script'],
+            ['style'],
+            ['textarea'],
+            ['title'],
+            ['xmp'],
```
nolanlawson, line 125: I tried to find all the "raw text" tags I knew of, based on [the Firefox source code](https://searchfox.org/mozilla-central/rev/72d959d715d3f832328057600811c09ee5195ae1/parser/html/nsHtml5Tokenizer.cpp#54-87).

Again, all these tests were already passing except for `noframes`.

fb55:
Thanks for this @nolanlawson! This also would be a good one to upstream to https://github.com/html5lib/html5lib-tests, which is used to the the majority of spec-compliant HTML parsers.

wooorm (approved):

43081j:
are we happy to merge this meanwhile? and open a separate issue to make sure we do add it to the upstream tests eventually

fb55:
@43081j Happy to do so

PR #1313, opened 2024-10-27
fix: export tree adapter types as namespace to support TS4.x

### LINKED ISSUE #1299: 7.2.0 throws a TS error TS1383: Only named exports may use 'export type'
The exact error is:

```
node_modules/parse5/dist/index.d.ts(5,1): error TS1383: Only named exports may use 'export type'.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

My temporary solution was to create a package.json resolutions entry for 7.1.2 since this package is a dependency of another package and not a direct dependency. That was after upgrading Node to 21 and TS to 5.4.3.

43081j:
Exports the tree adapter types as a namespace, individually. This reintroduces typescript 4.x support, which we accidentally broke in a non-breaking release.

Fixes #1299 

Before we merge this I need to double check it really does fix the problem. I tried building it with 4.x and it works but I need to check the types are still the same

### DIFF
--- a/packages/parse5/lib/index.ts
+++ b/packages/parse5/lib/index.ts
@@ -4,7 +4,21 @@ import type { DefaultTreeAdapterMap } from './tree-adapters/default.js';
 import type { TreeAdapterTypeMap } from './tree-adapters/interface.js';
 
 export { type DefaultTreeAdapterMap, defaultTreeAdapter } from './tree-adapters/default.js';
-export type * as DefaultTreeAdapterTypes from './tree-adapters/default.js';
+import type * as DefaultTreeAdapter from './tree-adapters/default.js';
+// eslint-disable-next-line @typescript-eslint/no-namespace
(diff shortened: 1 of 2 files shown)


43081j:
seems to fix it so I'm going ahead and merging this 👀

wooorm:
If you want a review, I’d recommend keeping PRs open some non-weekend days too. I personally try not to work on the weekends. Some places (rust, node, unifiedjs) use a lazy consensus time of 72 hours.
That being said, if you don’t believe a review is needed, I’m fine with you merging immediately

43081j:
i wasn't planning on waiting for a review for this one unless someone happened to be around

i already tested that the types work as expected and discussed it somewhat on the TS discord, so figured its probably all good

will keep it in mind though, ive lost track of days a lot when travelling lately, my bad.

Issue #314, opened 2020-04-10
The tree adapter is not used for updating the end location of an element


DMartens:
Hi,
I am currently writing a custom tree adapter restructuring the source code location:
```javascript
{
	start: { offset: 0, line: 1, column: 1 },
	end:   { offset: 0, line: 1, column: 1 },
}
```
The problem is that elements closing tag are not correctly handled in the [parser mixin](https://github.com/inikulin/parse5/blob/252819607421a5741cf745bb60c404f023531b0d/packages/parse5/lib/extensions/location-info/parser-mixin.js#L35).
The location is retrieved using the tree adapter but the end location is set directly on the retrieved location, not using the tree adapter.
Therefore the result after parsing is:
```javascript
{
	start: { offset: 0, line: 1, column: 1 },
	end:   { offset: 0, line: 1, column: 1 },
	endTag: ...,
    endLine: 1,
	endCol: 10,
	endOffset: 9,
}
```

I think there are three possible solutions:
1. Call the `setSourceCodeLocation` only after the closing tag is parsed
2. Call the `setSourceCodeLocation` twice (once after the opening tag and once for the closing tag)
3. Introduce a new method `updateSourceCodeLocation` for this purpose

Personally I would prefer the last option.
If this change is approved, I gladly submit a PR for the proposed solution.

inikulin:
Hi,

The original idea was that user don't have control over the source location structure. Therefore, `set/getSourceCodeLocation` methods should be used solely for data storage/retrievement and keep location data itself intact. 

This design was introduced because basically I didn't expect that someone would need custom location data representation. However, if it's a case for you, I'm not oppose to introduction of `updateSourceCodeLocation` and would welcome a PR for that. Though, not that would be a breaking change and will require a major version bump and proper documentation.

inikulin:
Closed in #315

PR #1366, opened 2025-01-06
chore: enable no-explicit-any rule as error level

43081j:
Removes our override such that `any` is no longer allowed, and replaces usages of it.

@fb55 this uncovered a curiosity:

the `Updating node source code location (GH-314)` test was basically asserting that you can have a custom source code location setter. i.e. you can have your own location shape

however, we have a strong type for `Location` and all tree adapters are required to return an `ElementLocation` from `getSourceCodeLocation` (which itself is a `Location`).

so while we do let you set whatever you want in `setSourceCodeLocation`, it doesn't make much sense to set anything that isn't a `Location`, as you need to get that back out in `getSourceCodeLocation` anyway.

tldr i think the test and the use case made sense way back, but has lost its meaning as we've strengthened types

### DIFF
--- a/packages/parse5-html-rewriting-stream/lib/index.ts
+++ b/packages/parse5-html-rewriting-stream/lib/index.ts
@@ -173,5 +173,6 @@ export interface RewritingStream {
      * @param event Name of the event
      * @param handler Event handler
      */
-    on(event: string, handler: (...args: never[]) => void): this;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    on(event: string, handler: (...args: any[]) => void): this;
 }
--- a/packages/parse5-html-rewriting-stream/lib/index.ts
+++ b/packages/parse5-html-rewriting-stream/lib/index.ts
@@ -173,5 +173,5 @@ export interface RewritingStream {
      * @param event Name of the event
      * @param handler Event handler
      */
-    on(event: string, handler: (...args: any[]) => void): this;
+    on(event: string, handler: (...args: never[]) => void): this;
 }
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -60,7 +60,6 @@ export default [
         rules: {
             '@typescript-eslint/no-unsafe-declaration-merging': 'off',
             '@typescript-eslint/no-non-null-assertion': 'warn',
-            '@typescript-eslint/no-explicit-any': 'warn',
             '@typescript-eslint/explicit-function-return-type': 'error',
             '@typescript-eslint/consistent-type-imports': 'error',
 
--- a/packages/parse5-parser-stream/lib/index.ts
+++ b/packages/parse5-parser-stream/lib/index.ts
@@ -144,5 +144,6 @@ export interface ParserStream<T extends TreeAdapterTypeMap = DefaultTreeAdapterM
      * @param event Name of the event
      * @param handler Event handler
      */
-    on(event: string, handler: (...args: never[]) => void): this;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    on(event: string, handler: (...args: any[]) => void): this;
 }
--- a/packages/parse5-parser-stream/lib/index.ts
+++ b/packages/parse5-parser-stream/lib/index.ts
@@ -144,5 +144,5 @@ export interface ParserStream<T extends TreeAdapterTypeMap = DefaultTreeAdapterM
      * @param event Name of the event
      * @param handler Event handler
      */
-    on(event: string, handler: (...args: any[]) => void): this;
+    on(event: string, handler: (...args: never[]) => void): this;
 }
--- a/packages/parse5-sax-parser/lib/index.ts
+++ b/packages/parse5-sax-parser/lib/index.ts
@@ -291,5 +291,5 @@ export interface SAXParser {
      * @param event Name of the event
      * @param handler Event handler
      */
-    on(event: string, handler: (...args: any[]) => void): this;
+    on(event: string, handler: (...args: never[]) => void): this;
 }
--- a/packages/parse5-sax-parser/lib/index.ts
+++ b/packages/parse5-sax-parser/lib/index.ts
@@ -291,5 +291,6 @@ export interface SAXParser {
      * @param event Name of the event
      * @param handler Event handler
      */
-    on(event: string, handler: (...args: never[]) => void): this;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    on(event: string, handler: (...args: any[]) => void): this;
 }
--- a/packages/parse5/lib/parser/parser-location-info.test.ts
+++ b/packages/parse5/lib/parser/parser-location-info.test.ts
@@ -1,6 +1,13 @@
 import * as assert from 'node:assert';
 import { outdent } from 'outdent';
-import { type ParserOptions, type TreeAdapterTypeMap, parse, parseFragment } from 'parse5';
+import {
+    type ParserOptions,
+    type TreeAdapterTypeMap,
+    parse,
+    parseFragment,
+    type DefaultTreeAdapterMap,
+    type Token,
+} from 'parse5';
 import {
     generateLocationInfoParserTests,
     assertStartTagLocation,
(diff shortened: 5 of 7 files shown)


fb55 (approved):
Very nice!

packages/parse5-html-rewriting-stream/lib/index.ts
```diff
@@ -173,5 +173,5 @@ export interface RewritingStream {
      * @param event Name of the event
      * @param handler Event handler
      */
-    on(event: string, handler: (...args: any[]) => void): this;
```
wooorm, line 176: if you’d want a type that is an array but without “values” in it, perhaps you want the empty tuple? `[]`

wooorm: Perhaps you are looking for `unknown`?

wooorm (commented):
As for “so while we do let you set whatever you want in setSourceCodeLocation, it doesn't make much sense to set anything that isn't a Location, as you need to get that back out in getSourceCodeLocation anyway.”

This project predates typescript, and also not every javascript user uses typescript. So would an alternative approach also be possible: to make the types match javascript?

packages/parse5-html-rewriting-stream/lib/index.ts
43081j: i'm not looking for `unknown` or trying to represent an empty set. i'm representing "any arguments":

```ts
type Fn = (...args: never[]) => void;

const fn0: Fn = (a: number): void => {}; // works
const fn1: Fn = (...args: unknown[]): void => {}; // works
const fn2: Fn = (): void => {}; // works

const fn3 = (foo: string, bar: number): void => {}
const fn4: Fn = fn3; // works
```

`any` is just disabling the type checker pretty much. the equivalent here is `never` since it is assignable to all types, whereas the opposite is true for `unknown` (all types are assignable to `unknown`)

in the example above, if you use `unknown[]`, you'll see an error for `fn0` as you can't assign `unknown` to `number` (which is what happens if you annotate the type as `Fn` since it expects the function you give it can handle `unknown[]`)

wooorm: If you want to represent any, then I recommend using `any`, and turning the lint rule off for these lines.

And, likely, a to do comment might also be in order. To type this untyped behavior. Or perhaps we could even remove it in the future. Why are unknown events allowed like this?

43081j: it isn't the same

`any` is effectively disabling the type checker for this type

`never[]` correctly defines what we're trying to define here, whether you're a fan of the name or not

in a world without the `any` escape hatch, this would indeed be `never[]`

wooorm: The problems:
- parse5 has untyped behavior, we allow arbitrary untyped event handlers
- `any` exists and is good, linting against it is also good, but it is good to configure linting rules

I do not think `never` is a good substitute

43081j: i'm a bit lost as this type is specifically to allow the arbitrary event handlers you speak of

```ts
interface Emitter {
    on(ev: 'strong-event', handler: () => void): void;
    on(ev: string, handler: (...args: never[]) => void): void;
}

declare const emitter: Emitter;

// error
emitter.on('strong-event', (foo: number) => {
});
```

maybe you are getting held up on the fact it is called `never`? as this is actually the correct type as far as i understand

43081j: As per the tseslint docs:

>  Using any disables many type checking rules and is generally best used only as a last resort or when prototyping code

because `any` is effectively disabling the type system, it isn't representing anything. its us saying "we don't want typescript here because we don't know how to type it"

the actual type of what we're trying to represent is "a function with any arguments", which is exactly what `...args: never[]` does (and only that)

i think this is because you're seeing `any` as a type like any other (thanks to the name), whereas i'm seeing it as "turn the type checker off" (which is how tseslint sees it)

wooorm: > its us saying "we don't want typescript here because we don't know how to type it"

I do not think that that is what `any` is — that is close to what `ts-ignore`/`ts-expect-error` are — `any` is an important part of typescript.

> the actual type of what we're trying to represent is "a function with any arguments", which is exactly what ...args: never[] does (and only that)

If we’d want any arguments, which is indeed a sometimes legitimate case, then that is what `any` is for: `...args: any[]`

Perhaps, as my words are not enough, you might find it interesting to look through `typescript/lib/*` in your `node_modules`. You will find that typescript itself often uses `any[]`. But never `never[]`.
You could also search for [`Andarist`](https://github.com/search?q=owner%3AAndarist+any%5B%5D+NOT+is%3Afork&type=code), or `mattpocock` for the difference. You will see that `any[]` is very common and `never[]` rarely used.
Lastly, if `never[]` was better than `any[]`, would [`no-explicit-any`](https://typescript-eslint.io/rules/no-explicit-any/) not suggest that then?

43081j: > I do not think that that is what any is — that is close to what ts-ignore/ts-expect-error are — any is an important part of typescript.

it is an important part of typescript. I quoted TSESLint directly, but it is fine that you disagree with it.

> If we’d want any arguments, which is indeed a sometimes legitimate case, then that is what any is for: ...args: any[]

i disagree. if we wanted any arguments, we want something you can assign a function with any arguments to, and that is what `never[]` is achieving here

> Perhaps, as my words are not enough, you might find it interesting to look through typescript/lib/* in your node_modules. 

no need.

> You could also search for Andarist, or mattpocock for the difference. You will see that any[] is very common and never[] rarely used.

i could ask the typescript team themselves too, but that's beyond the scope of this PR.

`any[]` is common, `never[]` is rarely used. but much of that is because `never` is a difficult concept to grasp

> Lastly, if never[] was better than any[], would [no-explicit-any](https://typescript-eslint.io/rules/no-explicit-any/) not suggest that then?

no because, as mentioned above, it is a difficult concept to grasp. it is easier to tell people to use `any` (mostly because of the naming, it makes more sense in peoples' heads).

> any / unknown would be the correct type here. I would lean towards any, so the user can pass any function themselves. This is also most commonly used in such situations indeed.

i disagree. i think we have the correct type, but the commonly used type in this situation is `any`. and _that_ is why we would use `any`

fully aware of why the rule exists too and that it isn't meant to be "don't ever use any"

---

this has gone on too long and isn't worth the time anymore. i am going to change back to `any` and move on to something else

43081j:
> This project predates typescript, and also not every javascript user uses typescript. So would an alternative approach also be possible: to make the types match javascript?

the parser assumes `getNodeSourceCodeLocation` returns an `ElementLocation`. this is concrete whether you use types or not

similarly, update/set are passed an `ElementLocation`

these are not in question and have the right types. however, `sourceCodeLocation` could technically be anything (on a node)

so, for this test to work as it used to, it'd need its own `Node` type which has a different `sourceCodeLocation`

packages/parse5-html-rewriting-stream/lib/index.ts
43081j: fwiw i asked about this in the typescript discord and this is what i was trying to explain to you:

https://github.com/microsoft/TypeScript/pull/55714/files

yes, you can find `any` in your node_modules, because it existed before `never` and `unknown` did

this PR will land and many others will too, leading to typescript itself doing exactly what i was trying to do here

similarly a very helpful quote:

> any shouldn't be read as "any". if TS were invented today with its current type system, i think any would be a different keyword (unsafe or unchecked or somesuch maybe)

43081j: @wooorm given typescript itself is moving to what i originally attempted to do (using `never`), does your opinion still hold?

if it does, i will change to `any` for now. just tell me one way or the other and i can get this merged

wooorm: Where does typescript do that?

wooorm: I do not draw that conclusion from that PR.

I also do not think that `no-explicit-any` wants `never`

43081j: interesting. they're changing the arguments to `never` on all the types which represent "some function", regardless of why the PR exists.

you said you saw them using `any` in your node_modules, but this is mostly because `never` didn't exist at the time. this PR will land and will change most of those occurrences (to do with functions at least) to use `never`

`Parameters<T>` is to get the parameters of a _function_, any function. so `T` must be constrained to "any function". In that PR, they choose to represent "any function" using `never` for the args.

meanwhile, we also want to represent "any function". so i did exactly what typescript itself soon will do

typescript itself is doing what i want to do, and many people who know better than you and i told me as much on their discord (to use `never` here).

if that isn't enough, we're never going to agree. so i'm going to use `any` and merge this as it is becoming a time sink

wooorm: > they're changing

I don’t have something to draw that conclusion from.

> because never didn't exist at the time

I’m not sure that is the reason — I think the more logical conclusion that people use `any` is because `any` is useful.

> In that PR, they choose to represent "any function" using never for the args.

It’s a stalled PR that does logic in the type system but doesn’t work. I do not deduce from that that `never` should be used in our case

> meanwhile, we also want to represent "any function". so i did exactly what typescript itself soon will do

I don’t think that’s what’s the goal here. I think the goal, per the PR title, is to enable a lint rule that `any` should not be used. The docs for that rule have more explanation for what it does and why one should use it: https://typescript-eslint.io/rules/no-explicit-any/. The alternatives it talks about are to actually type things. I don’t see it suggesting `never`. As I stated above, I think we do not need this arbitrary unsafe function support, and can also remove this functionality. But, meanwhile, `any` is useful, so we can also use `any`, with an ignore comment for now.

> typescript itself is doing what i want to do, and many people who know better than you and i told me as much on their discord (to use never here).

I don’t see typescript itself doing that, and I have not seen people say those things.

43081j: we're not going to agree

will put `any` back and merge 👍

PR #1376, opened 2025-01-21
chore: migrate to vitest

43081j:
I'd like to figure out how to use `include` to select coverage rather than having to individually `exclude` everything before merging. but it isn't a blocker

possible alternative to #1268 

vitest seems to produce better coverage somehow. i couldn't figure out why mocha/c8 was falling over

### DIFF
--- a/vitest.config.js
+++ b/vitest.config.js
@@ -1,0 +1,13 @@
+/// <reference types="vitest/config" />
+import { defineConfig } from 'vite';
+
+export default defineConfig({
+    test: {
+        include: ['**/*.test.ts'],
+        workspace: ['packages/*'],
+        coverage: {
+            clean: true,
+            exclude: ['./packages/*/dist', './scripts', './bench', './*.config.js'],
+        },
+    },
+});
--- a/vitest.config.js
+++ b/vitest.config.js
@@ -1,13 +1,29 @@
 /// <reference types="vitest/config" />
 import { defineConfig } from 'vite';
+import path from 'node:path';
+import { readdirSync } from 'node:fs';
+
+const alias = [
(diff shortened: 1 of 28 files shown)


wooorm (approved):

vitest.config.js
```diff
@@ -0,0 +1,13 @@
...
+export default defineConfig({
+    test: {
+        include: ['**/*.test.ts'],
+        workspace: ['packages/*'],
+        coverage: {
+            clean: true,
+            exclude: ['./packages/*/dist', './scripts', './bench', './*.config.js'],
+        },
+    },
+});
```
AriPerkkio, line 13: Missing Jest's remapping configs:

https://github.com/inikulin/parse5/blob/2094d5b08d11feda3f164bcf3e1e6aa65a5d0733/package.json#L63-L70

```suggestion
import path from 'node:path';
import { defineConfig } from 'vitest/config';

export default defineConfig({
    test: {
        include: ['**/*.test.ts'],

        // https://github.com/inikulin/parse5/blob/2094d5b08d11feda3f164bcf3e1e6aa65a5d,0733/package.json#L63C1-L70C11
        alias: [
            { find: /^(parse5[^/]*)\/dist\/(.*?)(?:\.js)?$/, replacement: path.resolve('./packages/$1/lib/$2') },
            { find: /^(parse5[^/]*)$/, replacement: path.resolve('./packages/$1/lib/index.ts') },
            { find: /parse5-test-utils\/(.*?)/, replacement: path.resolve('./test/$1') },
        ],

        coverage: {
            include: ['packages'],
            exclude: ['**/dist/**'],
        },
    },
});
```

43081j: have used your other suggestion from discord 👍

fb55 (approved):
Very nice!

PR #1410, opened 2025-02-20
chore(deps): bump entities from 4.5.0 to 6.0.0

cardoso:
Closes #1296

PR #1411, opened 2025-02-20
feat: switch to ESM-only

43081j:
Switches the 3 remaining dual packages to ESM only:

- `parse5-htmlparser2-tree-adapter`
- `parse5-parser-stream`
- `parse5`

**NOTE:** we still haven't decided if we are ready to do this yet. just leaving this draft here for when we do one day

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,8 +25,7 @@
         "vitest": "^3.0.1"
     },
     "scripts": {
-        "build": "npm run build:esm && npm run build:cjs --workspaces --if-present",
-        "build:esm": "tsc --build packages/* test",
+        "build": "tsc --build packages/* test",
         "build:docs": "typedoc",
         "prettier": "prettier '**/*.{js,ts,md,json,yml}' --log-level warn",
         "format": "npm run format:es && npm run format:prettier",
--- a/packages/parse5-html-rewriting-stream/package.json
+++ b/packages/parse5-html-rewriting-stream/package.json
@@ -20,7 +20,11 @@
     "main": "dist/index.js",
     "module": "dist/index.js",
     "types": "dist/index.d.ts",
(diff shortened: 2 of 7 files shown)


43081j:
@fb55 if i can get a thumbs up from you too, ill merge and publish this under a new major version

fb55:
I am unable to go ESM-only with Cheerio right now, so would prefer to keep the CJS build

43081j:
> I am unable to go ESM-only with Cheerio right now, so would prefer to keep the CJS build

are you on node `<20`?

also, unless you see us adding features, you could just stick to the current version, no? it seems unlikely we'd change API, so won't have anything in the new major that the old one doesn't have already

43081j:
@fb55 can we discuss this?

I'm in the process of migrating a lot of other packages I maintain to esm-only. so it'd be good to wrap this one up one way or another

if you still need CJS, that's what the old version is for since its unlikely we'll add any features any time soon. if somehow there is a fix, we can backport it if needed

cardoso:
Also https://github.com/inikulin/parse5/pull/1410 is now unblocked and could land in the current major, so in addition to feature parity, the dependencies won't stay behind.

43081j:
@fb55 maybe lets revisit this?

my plan would be this now:

1. fix the `entities` problem (not sure how yet, but do it in a way that its still dual-package but can be used without `node16` resolution in typescript)
2. release `entities` as ESM only (new major)
3. release `parse5` as ESM only (new major)

43081j:
we actually fixed the entities problem and the other internal issue without having to change esm/CJS in the end

so this should be unblocked now!

if I can get the go ahead from you @fb55 , I'll catch this up and sort out a new major etc

PR #1435, opened 2025-03-16
fix: handle exhaustive stack popping

43081j:
In some weird cases it seems you can end up trying to pop the element
stack until you basically reach `-1` (i.e. `indexOf` failed to find what
we want).

If that happens, the current element will be `undefined`. This change
adds some handling around that.

@fb55 this is the one we talked about separately

finding it difficult to write a test for it since its a really strange state to get into


43081j:
nevermind, it seems ill have to fix this another way as there are tests depending on it being `undefined`... 🤦‍♂️

43081j:
@wooorm @fb55 any chance you can both take a look at this or the other mentioned pr? Then whichever we're all happy with, we can merge

PR #1436, opened 2025-03-16
fix: account for undefined in element stack

43081j:
Changes the element stack such that the current element and ID can be
`undefined`, as this is already true but not represented in the types.

Also then updates any usages to account for this.

Depends on #1435 - it is basically that PR but with all other usages and the types fixed

### DIFF
--- a/packages/parse5/lib/parser/index.ts
+++ b/packages/parse5/lib/parser/index.ts
@@ -303,7 +303,7 @@ export class Parser<T extends TreeAdapterTypeMap> implements TokenHandler, Stack
         }
     }
 
-    protected _setContextModes(current: T['parentNode'], tid: number): void {
+    protected _setContextModes(current: T['parentNode'] | undefined, tid: number | undefined): void {
         const isHTML = current === this.document || (current && this.treeAdapter.getNamespaceURI(current) === NS.HTML);
 
         this.currentNotInHTML = !isHTML;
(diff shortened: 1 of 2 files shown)


fb55 (approved):
Very nice, let's go ahead with this?

packages/parse5/lib/parser/index.ts
```diff
@@ -419,7 +420,7 @@ export class Parser<T extends TreeAdapterTypeMap> implements TokenHandler, Stack
         } else {
             const parent = this.openElements.currentTmplContentOrNode;
 
-            this.treeAdapter.appendChild(parent, element);
+            this.treeAdapter.appendChild(parent ?? this.document, element);
```
wooorm, line 423: This one looks a bit iffy to me, to append to the entire document? I would *guess* something like `if (parent) { this.treeAdapter(parent, element) }` makes more sense?
But I am missing the likely needed context, so if you think it’s :+1:, then I am :+1:

(this applies to both PRs I guess)

(otherwise :+1:)

43081j: we could, its upto us what behaviour we want

basically, its possible to end up with a messy element stack in some rare cases, with broken syntax etc.

in those cases, the parent doesn't necessarily exist here. so we can either do nothing, or append to the root. if we do nothing, it may be confusing that nothing happened (empty tree). but I understand appending to the document may also be a surprise

I don't think the tests/spec define what to do here.

43081j: @wooorm what do you think we should do about this?

43081j: i'm not worried

i was just checking since i had written a response and not heard anything since

PR #1545, opened 2025-07-09
feat: bump to 8.0.0

43081j:
Bumps all packages and their parse5 dependencies to 8.0.0.

This isn't strictly necessary as some packages will continue to work
with 7.x, but it may be easier for maintenance in future to draw the
line.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,28 +39,6 @@
                 "parse5": "npm:parse5"
             }
         },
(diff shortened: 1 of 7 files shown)


dargmuesli:
@43081j I don't see any changelog update for this release, this may hinder adoption of v8, would you check out where it got lost? :eyes:

PR #1787, opened 2026-06-22
chore: use HTTPS repository URL

dangelo352:
Updates the package repository metadata to use HTTPS instead of the unauthenticated `git://` protocol.

Validation:
- `git diff --check`
- parsed the changed `package.json` file(s) with Node.js


43081j:
we're good thanks. happy to review when its not from an openclaw bounties repo 👍

PR #1842, opened 2026-09-02
feat: expose element creation context to tree adapters

domenic:
Add an optional onElementCreated tree-adapter callback that runs after element creation and before insertion, exposing the intended parent, form element pointer, and template-stack state. jsdom needs these parser-only values to implement [step 15 of “create an element for the token”](https://html.spec.whatwg.org/multipage/parsing.html#create-an-element-for-the-token), which establishes parser-created form owners, without subclassing the internal Parser class.

The parser computes each insertion location once so the callback context and subsequent insertion use the same target.

### DIFF
--- a/packages/parse5/lib/parser/index.ts
+++ b/packages/parse5/lib/parser/index.ts
@@ -28,6 +28,7 @@ import {
     type EOFToken,
     type LocationWithAttributes,
     type ElementLocation,
+    type Attribute,
 } from '../common/token.js';
 
 //Misc constants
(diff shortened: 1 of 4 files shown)


packages/parse5/lib/parser/index.ts
```diff
@@ -465,10 +493,11 @@ export class Parser<T extends TreeAdapterTypeMap> implements TokenHandler, Stack
 
     /** @protected */
     _insertFakeRootElement(): void {
-        const element = this.treeAdapter.createElement(TN.HTML, NS.HTML, []);
+        const intendedParent = this.openElements.current ?? this.document;
```
43081j: how come we now fall back to `this.document` explicitly? did we have a bug here? 🤔  (calling `appendChild(undefined, element)`)

domenic: Looks like over-defensive AI code, sadly. Thanks for finding it. I made it perform a self-review and it will fix this and a couple other small issues it found with an upcoming commit.

domenic:
Gentle ping! I would love to improve jsdom's form-association machinery based on this.

43081j:
This looks good, I'm just away this week at ZurichJS. 

I'll give it another review when I'm back and aim to merge 👍

PR #1854, opened 2026-09-17
feat: allow flushing buffered character tokens

domenic:
I asked GPT-6 Astra Extra High for the minimal parse5 change that would allow us to implement `document.write()` correctly in jsdom. It seems like it's mostly additive. It spent most of its time on tests, which is a good sign, and in particular was particularly concerned with getting the character offsets right.

I asked it a few times

> Perform adversarial self-review. Is this a PR a human maintainer would be happy to receive?

and by the third round it seemed like it was happy with the results.

Thank you for the great library :)


43081j:
given its still a large change, can you explain what its for?

what exactly are you trying to do in JSDOM that needs it? some context would be good

domenic:
It's for [jsdom's proper parser-blocking scripts / `document.write()` implementation](https://github.com/jsdom/jsdom/pull/4356). This should log `"hello"`, but currently logs `""`:

```js
document.open();
document.write("<p>hello");
console.log(document.querySelector("p").textContent);
```

We need to flush parse5's buffered text before `document.write()` returns, without ending the input stream.

